PR #39, opened 2024-05-06
Create csv2rdf_by_RDFLib.py

candlecao:
This is an optional method to convert an CSV to RDF in a serialized "turtle' format using Python coding.


cantusdb/turtle/csv2rdf_by_RDFLib.py
```diff
@@ -0,0 +1,51 @@
...
+
+def convert_csv_to_turtle(csv_filename):
+    with open(csv_filename, 'r', encoding='utf-8') as csv_file:
+        g = Graph()
+        csv_reader = csv.DictReader(csv_file)
+        
+        chant_type = URIRef(prefix_uri["cantusDB"] + "chant")
+        
+        # Add predicate URIs
+        p_incipit = URIRef(prefix_uri["wdt"] + "P1922")
```
dchiller, line 21: Could these be constants defined outside of the function? Also, you could use f-strings here:

```python
P_INCIPIT = URIRef(f"{prefix_uri['wdt']}P1922")
```

dchiller:
Two other comments/questions:
- it looks like you are using an external dependency `rdflib` and I don't see anywhere in the current repository where that dependency is defined. If it is already defined somewhere and I am not seeing it, that's good. But if not, you should define it somewhere, probably a `requirements.txt` file or a `pyproject.toml` file so that someone else could run this script. If you want some help doing this or want to work together, let me know!
- I'm wondering what the goal of this script is? Are you expecting to run it on multiple different files? If so, I think having some other way to pass the file in and get the output would be preferable. For example, could you pass in the file as a command line argument? Should the output go to some file rather than be printed to the terminal? The answers to these depends on the purpose...happy to help think through these as well.

Yueqiao12Zhang:
Hi,
The pyproject.toml file is created with all the required libraries.
The goal of the script is to generate a turtle format rdf file given a CSV file as input. I agree with what you said about inputting the filename as a command line argument, which makes the script run efficiently. The last print line is also fixed. In the last version, that line does not print to the terminal since the output texts are serialized to a new turtle file.

dchiller (approved):

cantusdb/turtle/csv2rdf_by_RDFLib.py
```diff
@@ -0,0 +1,71 @@
...
+    
+    # a specific config file for the predicates
+    # TODO: maybe change this part to make a more customized way of getting the predicates configurations
+    ontology_f = open("ontology_dict.csv", 'r')
+    ontology_reader = csv.DictReader(ontology_f)
+    ontology_dict = {}
+    
+    # converting the reader dict to the predicate dictionary
+    for row in ontology_reader:
+        ontology_dict[row[list(row.keys())[0]]] = row[list(row.keys())[1]]
```
ahankinson, line 18: I have no idea what this line is doing. Do you have an example output? Or better yet, just change the format of your file to match what you need instead of trying to fit it all into a.CSV.

cantusdb/turtle/csv2rdf_by_RDFLib.py
```diff
@@ -0,0 +1,71 @@
...
+    # Convert each row to Turtle format and add it to the output
+    for row in csv_reader:
+        # the first column as the subject
+        absolute_uri = URIRef(row[list(row.keys())[0]])
+        g.add((absolute_uri, RDF.type, URIRef("http://www.wikidata.org/prop/direct/chant")))
+        
+        # extracting other informations
+        for k in row:
+            # if this key is the first column, skip
+            if k.__eq__(list(row.keys())[0]):
```
ahankinson, line 44: Why are you using `__eq__`? Just use `==`. And the DictReader gives you the ability to use this as a dict, so getting the first entry in a list of keys completely passes that.

cantusdb/turtle/csv2rdf_by_RDFLib.py
```diff
@@ -0,0 +1,71 @@
...
+            # the object might be an URI or a literal
+            if validators.url(row[k]):
+                obj = URIRef(row[k])
+            else:
+                obj = Literal(row[k])
+            
+            # add this triple (absolute uri, predicate, object) to the graph
+            g.add((absolute_uri, predicate, obj))    
+        
+    csv_file.close()
```
ahankinson, line 62: If you use a `with` block when you open the file you don’t have to explicitly close the file. It willl close automatically.

cantusdb/turtle/csv2rdf_by_RDFLib.py
```diff
@@ -0,0 +1,71 @@
...
+
+    csv_file = open(sys.argv[1], 'r', encoding='utf-8')
+    
+    csv_reader = csv.DictReader(csv_file)
+    ontology_dict = get_ontology_dict()
+
+    # Convert each row to Turtle format and add it to the output
+    for row in csv_reader:
+        # the first column as the subject
+        absolute_uri = URIRef(row[list(row.keys())[0]])
```
ahankinson, line 38: `row[“absolute_url”]`?

cantusdb/turtle/csv2rdf_by_RDFLib.py
Yueqiao12Zhang: Hi, here is an example input:
relation,URI
absolute_url,
incipit,http://www.wikidata.org/prop/direct/P1922
genre,https://schema.org/genre
genre_id_url,http://www.wikidata.org/prop/direct/P136
mode_name,https://schema.org/musicalKey
mode_id_url,https://www.wikidata.org/wiki/Q731978
finalis,https://www.wikidata.org/wiki/Q4484726
src_link,https://cantusdatabase.org/sources
cantus_id,https://cantusindex.org/id
indexing_notes,https://www.wikidata.org/wiki/Q263478

Goes to output

{'absolute_url': '', 
'incipit': 'http://www.wikidata.org/prop/direct/P1922', 
'genre': 'https://schema.org/genre', 
'genre_id_url': 'http://www.wikidata.org/prop/direct/P136', 
'mode_name': 'https://schema.org/musicalKey', 
'mode_id_url': 'https://www.wikidata.org/wiki/Q731978', 
'finalis': 'https://www.wikidata.org/wiki/Q4484726', 
'src_link': 'https://cantusdatabase.org/sources', 
'cantus_id': 'https://cantusindex.org/id', 
'indexing_notes': 'https://www.wikidata.org/wiki/Q263478'}

It changes an input CSV file into a dictionary line by line. I think there's a clearer way of doing this, I will figure it out today.

cantusdb/turtle/csv2rdf_by_RDFLib.py
Yueqiao12Zhang: This script should be more general. It uses whatever is the first attribute of an entry in the input CSV file. In this case, it is the row["absolute_url"], but it can be other.

cantusdb/turtle/csv2rdf_by_RDFLib.py
Yueqiao12Zhang: Ok, I will delete this line.

cantusdb/turtle/csv2rdf_by_RDFLib.py
Yueqiao12Zhang: The first entry should be the first entry in the dictionary. The first entry in the dictionary should be the key attribute of each entry in the CSV file, it's not the header file. The purpose of this if statement is because the absolute_uri key attribute is already extracted and should be the subject of each triple relation of this entry.

cantusdb/turtle/csv2rdf_by_RDFLib.py
ahankinson: I guess I'm suggesting creating a file called "mapping.json", or even better (since it's not very large) just keeping the mapping in your code. Then you actually don't need to convert from a CSV to the dictionary.

cantusdb/turtle/csv2rdf_by_RDFLib.py
ahankinson: `__eq__(self)` is the method that defines how the `==` operator compares two things. Overriding it allows you to specify how you want equality to be computed, and is generally reserved for when you have some complex data type where equality isn't straightforward. 

In this case, though, I think you're just comparing strings? So `k == [whatever]` does the exact same thing.

https://docs.python.org/3/library/operator.html#operator.__eq__

cantusdb/turtle/csv2rdf_by_RDFLib.py
ahankinson: You can't just delete the line without moving it to a `with` block. So:

```
csv_file = open(sys.argv[1], 'r', encoding='utf-8')
```

should become:

```
with open(sys.argv[1], 'r', encoding='utf-8') as csv_file:
    ... stuff goes here ... 
```

Then you can remove the explicit call to `close()` since it will automatically close when the scope is exited.

cantusdb/turtle/csv2rdf_by_RDFLib.py
ahankinson: If that's the case then you probably should be using a different CSV Reader class than the DictReader, since the ability to access columns by their header is the whole point of using the DictReader class.

cantusdb/turtle/csv2rdf_by_RDFLib.py
Yueqiao12Zhang: Yes, I forgot to mention this

cantusdb/turtle/csv2rdf_by_RDFLib.py
Yueqiao12Zhang: Yes, I will change this. I was used to java and had a little confusion.

cantusdb/turtle/csv2rdf_by_RDFLib.py
Yueqiao12Zhang: I will figure out this problem with the first conversation.

cantusdb/turtle/csv2rdf_by_RDFLib.py
ahankinson: > The first entry in the dictionary should be the key attribute of each entry in the CSV file, it's not the header file

In your CSV file the first row gives your column headers. When you open your CSV file with the DictReader, and do not supply a list of fieldnames, it will interpret the first row as your field names. Thus:

```csv
absolute_url,incipit,genre,genre_id_url,mode_name,mode_id_url,finalis,src_link,cantus_id,indexing_notes,source_description
https://cantusdatabase.org/chant/562633,Ecce nunc palam loqueris et,Antiphon,https://www.wikidata.org/wiki/Q582093,Hypomixolydian mode,https://www.wikidata.org/wiki/Q57976142,G,https://cantusdatabase.org/source/123756,002530,,
```

Becomes: 

```py
{"absolute_url": "https://cantusdatabase.org/chant/562633",
"incipit": "Ecce nunc palam loqueris et",
"genre": "Antiphon",
"genre_id_url": "https://www.wikidata.org/wiki/Q582093",
... etc. }
```

This is done for you so that when you iterate over each row you don't need to access the fields by index, but can simply use the column names in that row: `row["absolute_url"]` rather than `row[list(row.keys())[0]]`.

Yueqiao12Zhang: Not in all cases the key attribute is named "absolute_url". It depends on the mapping, but it is always going to be the first entry in the dictionary.

ahankinson: That sounds like a recipe for a real headache. :-D

PR #41, opened 2024-05-14
Add script to convert a reconciled csv to turtle RDF

Yueqiao12Zhang:
To import our databases to Virtuoso in RDF triples format, we plan to download raw files from the databases, convert them into CSV, reconcile them, convert them to RDF turtle, and import them into Virtuoso. 
This pull request consists of a Python script, a mapper .JSON file, and other related testing files. It converts a reconciled CSV file into an RDF graph file in turtle format, with the help of a mapper .JSON file.
The CSV2RDF python script is now done, with an example mapper file related to an example large CSV file reconciled from OpenRefine.


.DS_Store
dchiller, line 1: `.DS_Store` files are Mac-specific files related to the file system user interface. They shouldn't be committed...you could add them to a `.gitignore`.

cantusdb/turtle/relations_mapping.json
dchiller, line 1: What is the relationship between `relations_mapping.json` and `relations_mapping.csv`? Do we need them both?

cantusdb/turtle/csv2rdf_by_RDFLib.py
```diff
@@ -3,64 +3,60 @@
...
+                    obj = URIRef(row[i])
+                else:
+                    obj = Literal(row[i])
+                
+                # add this triple (absolute uri, predicate, object) to the graph
+                g.add((key_attribute, predicate, obj))    
+                
     return g
 
 # openrefine_csv_test.csv
```
dchiller, line 62: Is this comment necessary?

cantusdb/turtle/csv2rdf_by_RDFLib.py
dchiller, line 1: Could you add docstrings to the functions here?

For example, for `get_ontology_dict`:

```
def get_ontology_dict():
"""
Returns the ontology found in "relations_mapping.json" as
python objects.
"""
```

cantusdb/turtle/csv2rdf_by_RDFLib.py
```diff
@@ -3,64 +3,60 @@
...
     if len(sys.argv) != 2:
-        raise ValueError("Invalid number of input filename") 
-    
-    g = Graph()
-
-    csv_file = open(sys.argv[1], 'r', encoding='utf-8')
-    
-    csv_reader = csv.DictReader(csv_file)
-    ontology_dict = get_ontology_dict()
+            raise ValueError("Invalid number of input filename")
```
dchiller, line 19: What is the vision for how this utility is to be used? As a command line tool? If so, it would be good for the arguments to have help strings, checks on type, etc. (for example, take a look at python's argparse module). Is it as a python module? If so, then this function should take some arguments.

dchiller (commented):
A few other things to add about the PR as a whole:
- has this code been formatted by black? 
- what has been introduced/changed in this PR? Is it connected to an issue? It is a good idea to add in the PR comments a description of why the PR is necessary, what it adds or fixes in the codebase (in broad terms).

cantusdb/turtle/relations_mapping.json
Yueqiao12Zhang: We do not, the CSV has been used before, and it takes many unnecessary conversions in the script to parse it.
The JSON file is much more simpler.

cantusdb/turtle/csv2rdf_by_RDFLib.py
Yueqiao12Zhang: I will add them.

cantusdb/turtle/csv2rdf_by_RDFLib.py
```diff
@@ -3,64 +3,60 @@
...
         
-    csv_file.close()
-    
+        header = next(csv_reader)
+
+        # Convert each row to Turtle format and add it to the output
+        for row in csv_reader:
+            # the first column as the subject
+            key_attribute = URIRef(row[0])
+            if "type" in list(ontology_dict.keys()):
```
dchiller, line 35: If `type` is not in `ontology_dict` what happens to the first column of each row?  Is it just not added to the turtle file?

cantusdb/turtle/csv2rdf_by_RDFLib.py
Yueqiao12Zhang: In the future, it might be used as a command in some executable. I will add help strings and type checks.

cantusdb/turtle/csv2rdf_by_RDFLib.py
```diff
@@ -3,64 +3,60 @@
...
+                g.add((key_attribute, RDF.type, URIRef(ontology_dict["type"])))
+            
+            # extracting other informations
+            # TODO: is the source description texts needed?
+            for i in range(len(row)):
```
dchiller, line 40: I think you could simplify this `for` block by: 

1. using the `enumerate` function (this way you can iterate through row and get both the index and the value.
2. iterating through `row[1:]` if you don't want the first column

cantusdb/turtle/csv2rdf_by_RDFLib.py
Yueqiao12Zhang: Yes. The type attribute should be added to the mapper file. I am still considering what should be a consequence if the type is not in the mapper since it depends on the definitions and protocols of the TTL file.

cantusdb/turtle/csv2rdf_by_RDFLib.py
Yueqiao12Zhang: Ok. I will fix it.

cantusdb/turtle/relations_mapping.json
dchiller: Can you remove the `.csv` then?

cantusdb/turtle/csv2rdf_by_RDFLib.py
dchiller: I think one good check for a PR is whether a reviewer could test it... and in this case I'd need some more information either in the code or in documentation about what and how to pass in what needs to be passed.

cantusdb/turtle/relations_mapping.json
```diff
@@ -0,0 +1,14 @@
...
+    "incipit": "http://www.wikidata.org/prop/direct/P1922",
+    "genre":"https://schema.org/genre",
+    "genre_id_url":"http://www.wikidata.org/prop/direct/P136",
+    "mode_name":"https://schema.org/musicalKey",
+    "mode_id_url":"https://www.wikidata.org/wiki/Q731978",
+    "finalis":"https://www.wikidata.org/wiki/Q4484726",
+    "src_link":"https://cantusdatabase.org/sources",
+    "cantus_id":"https://cantusdatabase.org/about/id-numbers",
+    "indexing_notes":"https://cantusdatabase.org/description/#IndexingNotes",
+    "type":"http://www.wikidata.org/prop/direct/chant",
```
ahankinson, line 12: Perhaps https://www.wikidata.org/wiki/Q26473?

cantusdb/turtle/csv2rdf_by_RDFLib.py
```diff
@@ -3,64 +3,60 @@
 import csv
 import validators
 import sys
+import json
 
 # helper to get the predicate relation dictionary
 def get_ontology_dict () :
```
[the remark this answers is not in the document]
ahankinson: Also, you should be using an editor that will alert you to PEP8 violations, like the extra space between the function name and the parens.

cantusdb/turtle/csv2rdf_by_RDFLib.py
ahankinson, line 19: This indent level does not match the rest.

Yueqiao12Zhang:
Do you think that a predicate in an RDF triple should always be a property? Or can it be an item?

fujinaga:
Ideally, a predicate should be an Wikidata Property, but it may be an Wikidata Instance, if no appropriate property exists in Wikidata. This would be the responsibility of the human reconciler, why do you ask?

cantusdb/turtle/csv2rdf_by_RDFLib.py
ahankinson: Also, you're checking this on every iteration of the loop, when you should be able to check and raise the `ValueError` in one place. If it passes, then you know that your ontology dict is complete.

cantusdb/turtle/csv2rdf_by_RDFLib.py
```diff
@@ -3,69 +3,63 @@
...
-    csv_file.close()
-    
+                raise ValueError("No type specifications in the mapper file.")
+
+            # extracting other informations
+            # TODO: is the source description texts needed?
+            for i, element in enumerate(row[1:]):
+                predicate_cur = header[i]
+                # finding the predicate from csv in the config dictionary, if not exit, skip
+                if predicate_cur in list(ontology_dict.keys()):
```
ahankinson, line 44: Consider reversing this so that you don't need the `else` --

```
if predicate not in ...: 
    continue

predicate = URIRef(...)
```

Yueqiao12Zhang:
> Ideally, a predicate should be an Wikidata Property, but it may be an Wikidata Instance, if no appropriate property exists in Wikidata. This would be the responsibility of the human reconciler, why do you ask?

Since logically, I think using an Wikidata Property as a predicate makes much more sense, but many predicates exist as an Wikidata instance. I'm wondering if using an instance (or item) as a predicate would be incorrect. Thank you for answering.

fujinaga:
We need to discuss these carefully when Properties are not available as a predicate. Can you provide some examples?

cantusdb/turtle/csv2rdf_by_RDFLib.py
```diff
@@ -3,69 +3,63 @@
...
+                    predicate = URIRef(ontology_dict[predicate_cur])
+                else:
+                    continue
+
+                # the object might be an URI or a literal
+                if validators.url(element):
```
dchiller, line 50: I'm noticing in the csv file that columns are consistent in whether or not they are URI's. But you are checking this every time anew on a row by row basis? Is the current csv a particular case where that happens to be true? Or is that something we are expecting or enforcing?

fujinaga:
We probably should move this thread to Discussions, as it is not directly related to this issue.

cantusdb/turtle/csv2rdf_by_RDFLib.py
ahankinson: Also, just looking at this more closely... 

`if predicate_cur in list(ontology_dict.keys())` is the same as `if predicate_cur in ontology_dict`, except for the fact that `x in list` is `O(n)`, while `x in dict` is `O(1)` (searching a list vs. searching a hashmap). It's not a big deal here because the lists are so small, but it's a good habit to always know the most performant path. 

Slow code that you don't write is code you don't have to optimize later.

cantusdb/turtle/csv2rdf_by_RDFLib.py
Yueqiao12Zhang: Actually, in the small example CSV I used to test before, there are inconsistencies in whether the attribute is URI or not. But in this newly reconciled big CSV file, they are consistent. I also expect that in the future they will be consistent.

Yueqiao12Zhang:
I think a recursive approach with a path-finding algorithm (O(n)?) would be much faster than a nested loop algorithm (O(n^2)). I will try to apply it.

ahankinson:
I would suggest doing that if you find your import is taking days. If it’s running in the order of a few seconds, no need to overcomplicate it.

.gitignore
```diff
@@ -1,3 +1,7 @@
 *.pyc
 */__pycache__
-psqldump
\ No newline at end of file
+psqldump
+musicbrainz/.DS_Store
+/musicbrainz/data
```
dchiller, line 5: This directory is already in the repo...are you sure you want to ignore it?

.gitignore
```diff
@@ -1,3 +1,7 @@
 *.pyc
 */__pycache__
-psqldump
\ No newline at end of file
+psqldump
+musicbrainz/.DS_Store
```
dchiller, line 4: Why not add all .DS_Store here?

cantusdb/turtle/test.ttl
dchiller, line 1: Should this file be included?

cantusdb/turtle/csv2rdf_single_subject.py
```diff
@@ -0,0 +1,67 @@
+from rdflib import Graph, URIRef, Literal
+from rdflib.namespace import RDF
+import csv
```
dchiller, line 3: Import order is not PEP-compliant... not a huge deal, but something a linter would point out automatically. https://peps.python.org/pep-0008/#imports

cantusdb/turtle/csv2rdf_single_subject.py
```diff
@@ -0,0 +1,67 @@
...
+                    obj = URIRef(element)
+                else:
+                    obj = Literal(element)
+
+                g.add((key_attribute, predicate, obj))
+
+    return g
+
+
+# openrefine_csv_test.csv
```
dchiller, line 62: What is this comment saying? This file is not referenced in this script/function.

cantusdb/turtle/csv2rdf_single_subject.py
```diff
@@ -0,0 +1,67 @@
...
+        with open("relations_mapping.json", "r") as mapper:
+            ontology_dict = json.load(mapper)
+
+        header = next(csv_reader)
+
+        # Convert each row to Turtle format and add it to the output
+        for row in csv_reader:
+            # the first column as the subject
+            key_attribute = URIRef(row[0])
+            if "type" in ontology_dict:
```
dchiller, line 35: I think I mentioned this before, but I wonder why you are checking this for every row?

cantusdb/turtle/csv2rdf_single_subject.py
```diff
@@ -0,0 +1,67 @@
...
+                g.add((key_attribute, RDF.type, URIRef(ontology_dict["type"])))
+
+            # extracting other informations
+            # TODO: is the source description texts needed?
+            for i, element in enumerate(row[1:]):
+                predicate_cur = header[i + 1]
```
dchiller, line 41: As above, anything you are doing with the header I think could be done once, before iterating through all of the csv rows.

cantusdb/turtle/csv2rdf_single_subject.py
Yueqiao12Zhang: This is the test file used for this script. This comment is here to ease copy and paste.

cantusdb/turtle/csv2rdf_single_subject.py
Yueqiao12Zhang: It does not need to be checked for each row, since the ontology dict is not modified, but I am having difficulty figuring out how to make it check only once. Isn't it O(1)?

ahankinson: Add: 

`ontology_type = ontology_dict["type"]`

outside of the loop

Yueqiao12Zhang: > Add:
> 
> `ontology_type = ontology_dict["type"]`
> 
> outside of the loop

It adds the s->type->o into the graph for each subject in the csv if the key "type" exists in the dict. If I add this line, how does it avoid adding the triple with a null predicate into the graph? Since the key "type" is not always present.

ahankinson: Your `ontology_dict["type"]` lookup ALWAYS returns this:

https://github.com/DDMAL/linkedmusic-datalake/pull/41/files#diff-0e440507c8aa12b4cc65117dae6b4b07412981bf7f03dfb63b03847f05aa0083R3

It does not change with each iteration of the loop. If you intended something different, then this is a bug.

Yueqiao12Zhang: Yes, but it might not exist in the dictionary json file. I might return null, and it's not a bug.

Yueqiao12Zhang: If it does not exist, no type predicate will be added to all the subjects.

ahankinson: It won't return null, since if it doesn't exist in the dictionary file it will raise a `KeyError` exception and your code will crash because you are trying to access a value that does not exist. 

```
>>> x = {'foo': 'bar'}
>>> x['type']
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyError: 'type'
```

Yueqiao12Zhang: Yes, that's why I check if type exists in that if statement.

ahankinson: *sigh* You do you, man.

ahankinson: I understand very well. I just think it's wrong, and I don't have a lot of time to try and convince you otherwise.

You are checking a value that will not change, because your values are hard-coded: https://github.com/DDMAL/linkedmusic-datalake/pull/41/files#diff-afb01fca9a39deafd6c3534c1fa395e76ce924535613fc59b69a77381aee9abbR26-R27

**If the type value does not exist, then you don't need to check it at every iteration of your loop -- you know it does not exist before you enter the loop**. This data is not modified once it is read. It does not change with each iteration.

ahankinson:
```
with open("relations_mapping.json", "r") as mapper:
     ontology_dict = json.load(mapper)

header = next(csv_reader)
ontology_type: Optional[str] = ontology_dict.get("type"}

# Convert each row to Turtle format and add it to the output
for row in csv_reader:
    # the first column as the subject
    key_attribute = URIRef(row[0])
    if ontology_type:
        g.add((key_attribute, RDF.type, URIRef(ontology_type)))
```

ahankinson: That's a quick-and-dirty example of what I'm trying to say. It's not a big deal.

ahankinson: Is computational complexity the only important thing to consider when writing code?

.gitignore
dchiller: > Since I'm working on another branch at the same time, how can I avoid having to choose the files to commit each time without adding these to the .gitignore?

I'm not sure exactly what the problem you are running into, but a temporary commit or `git stash` might do it for you. Choosing which files to include in a commit feels like very common practice to me, and certainly better than adding something to gitignore unless you actually intend to ignore the files...

cantusdb/turtle/csv2rdf_single_subject.py
dchiller: But as another developer, who only has access to this repository and not any other files you only have on your local machine, I don't know what `openrefine_csv_test.csv` is. What do I need to run this script? A csv created through openrefine? If so, then I think that's what your doctring/argument help-text should say.

cantusdb/turtle/csv2rdf_single_subject.py
dchiller: For every row in the csv, you are:
- slicing the `header` list to obtain the value of the header for that row (calling this `predicate_cur`)
- looking for whether or not `predicate_cur` is in `ontology_dict` (line 49)
- creating a URIRef based on that `predicate_cur` (line 52)

All of those steps are they exact same for each row, right? 

For every row, `header[i]` will be the same, so for every row, `header[i]` will always be in `ontology_dict` or it won't, and so the URIRef you create for a specific column will be the same for every row. 

So then I wonder why you are having your code do the same thing over and over again.

dchiller:
FYI: I am going to start resolving some of these discussions just so it is easier for me to see where things are...

.DS_Store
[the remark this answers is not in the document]
Yueqiao12Zhang: I added it in .gitignore.

cantusdb/turtle/csv2rdf_single_subject.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+# Run this file in the command line.
+# Arg 1 = the input reconciled csv using openrefine
+# An example in this directory: openrefine_csv_test.csv
+# is generated by openrefine and used for testing
+# Convert the CSV data to Turtle format
+if __name__ == "__main__":
+    turtle_data = convert_csv_to_turtle(sys.argv[1])
+    turtle_data.serialize(format="turtle", destination="test.ttl")
```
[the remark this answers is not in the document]
Yueqiao12Zhang: No, I will fix it.

cantusdb/turtle/csv2rdf_single_subject.py
```diff
@@ -0,0 +1,71 @@
...
+                predicates.append(URIRef(ontology_dict[column]))
+            
+        if "type" in ontology_dict:
+            ontology_type: Optional[str] = ontology_dict["type"]
+        else:
+            ontology_type = None
+            
+        # Convert each row to Turtle format and add it to the output
+        for row in csv_reader:
+            # the first column as the subject
```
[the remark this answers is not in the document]
Yueqiao12Zhang: I will add this in the docstring.

cantusdb/turtle/csv2rdf_single_subject.py
```diff
@@ -0,0 +1,71 @@
...
+    if len(sys.argv) != 2:
+        raise ValueError("Invalid number of input filename")
+
+    with open(filename, "r", encoding="utf-8") as csv_file:
+        # TODO: maybe change this one to accomadate for multiple files
+
+        g = Graph()
+
+        csv_reader = csv.reader(csv_file)
+        with open("relations_mapping.json", "r") as mapper:
```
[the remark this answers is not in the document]
Yueqiao12Zhang: This should be an absolute path.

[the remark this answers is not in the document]
Yueqiao12Zhang: Fixed!

cantusdb/turtle/csv2rdf_single_subject.py
```diff
@@ -0,0 +1,71 @@
...
+
+# a main operation function
+def convert_csv_to_turtle(filename) -> Graph:
+    """
+    Adds all informations as RDF triples from the input filename into a graph and return it.
+    @Pre: input filename must be type string.
+    @Post: Returns a RDF.Graph that has all the triples
+    """
+
+    if len(sys.argv) != 2:
```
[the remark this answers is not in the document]
Yueqiao12Zhang: Fixed!

cantusdb/turtle/csv2rdf_single_subject.py
```diff
@@ -0,0 +1,71 @@
...
+        # Convert each row to Turtle format and add it to the output
+        for row in csv_reader:
+            # the first column as the subject
+            key_attribute = URIRef(row[0])
+            if ontology_type:
+                g.add((key_attribute, RDF.type, URIRef(ontology_dict["type"])))
+
+            # extracting other informations
+            for i, element in enumerate(row[1:]):
+                if element == "" :
```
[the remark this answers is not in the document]
Yueqiao12Zhang: Yes, the black formatter.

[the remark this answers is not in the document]
dchiller: Ok, looks like it ran here now.

cantusdb/turtle/csv2rdf_single_subject.py
```diff
@@ -0,0 +1,71 @@
...
+        # TODO: maybe change this one to accomadate for multiple files
+
+        g = Graph()
+
+        csv_reader = csv.reader(csv_file)
+        with open("relations_mapping.json", "r") as mapper:
+            ontology_dict = json.load(mapper)
+
+        header = next(csv_reader)
+        header = header[1:]
```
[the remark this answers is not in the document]
dchiller: A comment or descriptive variable name would be helpful then. So what about:
`header_without_subject = header[1:]` or
```
# Here we remove the first value in the header since the first column contains the subject
header = header[1:]
```

cantusdb/turtle/csv2rdf_single_subject.py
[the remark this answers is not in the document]
dchiller: I don't think you can make it an absolute path...my file system will be different than yours. I'm saying this kind of requirement would be good to document somewhere.

cantusdb/turtle/csv2rdf_single_subject.py
```diff
@@ -0,0 +1,71 @@
...
+from rdflib import Graph, URIRef, Literal
+from rdflib.namespace import RDF
+from typing import Optional
+
+# a main operation function
+def convert_csv_to_turtle(filename) -> Graph:
+    """
+    Adds all informations as RDF triples from the input filename into a graph and return it.
+    @Pre: input filename must be type string.
+    @Post: Returns a RDF.Graph that has all the triples
```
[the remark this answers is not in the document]
dchiller: It's fine to have preconditions imho, but it's just that these don't look like that to me... they look like arguments and return values.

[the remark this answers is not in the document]
Yueqiao12Zhang: Ok. I will change it.

cantusdb/turtle/csv2rdf_single_subject.py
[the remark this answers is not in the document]
Yueqiao12Zhang: Yes. I'm wondering how can I track this mapper file based on the location of the script even if I am running it in other places.

cantusdb/turtle/csv2rdf_single_subject.py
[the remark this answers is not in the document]
Yueqiao12Zhang: Ok, I will fix it.

cantusdb/turtle/csv2rdf_single_subject.py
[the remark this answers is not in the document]
Yueqiao12Zhang: What was the mistake? I did not see any difference.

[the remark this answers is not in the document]
dchiller: A space between `""` and `:`

dchiller:
@Yueqiao12Zhang 

Feel free to resolve comments when you feel they are resolved and "Re-request review" from me (in the upper right, under "Reviewers") when you feel like the code is ready for another review.

.DS_Store
dchiller, line 1: Please remove this `.DS_Store` and others from the PR.

.gitignore
```diff
@@ -1,3 +1,6 @@
 *.pyc
 */__pycache__
-psqldump
\ No newline at end of file
+psqldump
+*/.DS_Store
+musicbrainz/*
```
dchiller, line 5: I'm still confused why you are ignoring everything in the `musicbrainz` folder!

cantusdb/turtle/csv2rdf_single_subject.py
```diff
@@ -0,0 +1,74 @@
...
+        ontology_dict = json.load(open(mappername, "r"))
+
+        header = next(csv_reader)
+        header_without_subject = header[1:]
+        predicates = []
+        for column in header_without_subject:
+            if column in ontology_dict:
+                predicates.append(URIRef(ontology_dict[column]))
+
+        ontology_type: Optional[str] = ontology_dict["type"]
```
dchiller, line 38: This looks like another requirement: the `ontology_dict` must contain a key of `type`. Please document this.

cantusdb/turtle/csv2rdf_single_subject.py
```diff
@@ -0,0 +1,74 @@
...
+import sys
+import json
+import os
+from rdflib import Graph, URIRef, Literal
+from rdflib.namespace import RDF
+from typing import Optional
+
+dirname = os.path.dirname(__file__)
+mappername = os.path.join(dirname, 'relations_mapping.json')
+destname = os.path.join(dirname, 'out_rdf.ttl')
```
dchiller, line 12: Are these meant to be configured or not?

dchiller (commented):
I left a few more comments in this review. Please also add the issue which this resolves or update the PR description with a description of the functionality that this PR introduces.

cantusdb/turtle/csv2rdf_single_subject.py
Yueqiao12Zhang: No, it does not have to contain "type". My previous implementation is: if there is no "type", then it is None.

ahankinson: If there is no type key, this code will raise a KeyError Exception. You want: 

```
 ontology_type: Optional[str] = ontology_dict.get("type")
```

That will return `None` if it does not have a type key.

cantusdb/turtle/csv2rdf_single_subject.py
```diff
@@ -0,0 +1,74 @@
...
+                predicates.append(URIRef(ontology_dict[column]))
+
+        ontology_type: Optional[str] = ontology_dict["type"]
+
+        # Convert each row to Turtle format and add it to the output
+        for row in csv_reader:
+            # the first column as the subject
+            key_attribute = URIRef(row[0])
+            if ontology_type:
+                g.add((key_attribute, RDF.type, URIRef(ontology_dict["type"])))
```
ahankinson, line 45:
```suggestion
                g.add((key_attribute, RDF.type, URIRef(ontology_type)))
```

.gitignore
Yueqiao12Zhang: I changed it before creating a new branch and I did not want the changes to reflect on this branch. I fixed it now.

.DS_Store
Yueqiao12Zhang: I fixed it.

cantusdb/turtle/csv2rdf_single_subject.py
dchiller: Ok, I think this is fine for now.

cantusdb/turtle/csv2rdf_single_subject.py
dchiller: Ok, @Yueqiao12Zhang 

If you fix this and if you provide a clearer description in the PR text (at the top of this page) and/or in a linked issue, I will approve.

dchiller:
"is done and ready to merge" is not really part of a descriptive PR title. You can designate that by re-requesting review or moving the PR from draft (this one is already not a draft, so not relevant here. A better PR name would be something descriptive about what it does. So, I've renamed this PR "Add script to convert a reconciled csv to turtle RDF"

cantusdb/turtle/csv2rdf_single_subject.py
dchiller, line 12:
```suggestion
DIRNAME = "."
mapping_filename = os.path.join(DIRNAME, 'relations_mapping.json')
dest_filename = os.path.join(DIRNAME, 'out_rdf.ttl')
```

You kind of re-engineered a more complicated relative file reference? I'm still not totally sure what your intention is. Do you want a user to be able to configure the file names and paths? 

This suggestion I guess shows a default to these file names in the current directory but then could be modified to not the current directory?

Yueqiao12Zhang: I think if you assign "." to DIRNAME, it directs to the current position where you are running the script. Only __file__ is where the file is located.

dchiller: True. So the desired behaviour here is that you can run this script from anywhere but that those files must be/will be in the directory where the script lives?

dchiller (approved):

cantusdb/turtle/csv2rdf_single_subject.py
Yueqiao12Zhang: Yes, since this script applies to the CSV files from other databases.

Issue #45, opened 2024-05-16
Importing MusicBrainz data and reconcile it using OpenRefine.


Yueqiao12Zhang:
Can you put a hold on these tasks?
I don't know what this ER schema is. You can explain this to me on Friday.

Can you work on importing some MusicBrainz data then reconcile it using OpenRefine? Make an issue for that.

_Originally posted by @fujinaga in https://github.com/DDMAL/linkedmusic-datalake/issues/43#issuecomment-2113071989_

Yueqiao12Zhang:
```
"artist-credit": [
            {
                "name": "Pigface",
                "joinphrase": "",
                "artist": {
                    "type-id": "e431f5f6-b5d2-343d-8b36-72607fffb74b",
                    "genres": [
                        {
                            "disambiguation": "",
                            "count": 1,
                            "id": "89255676-1f14-4dd8-bbad-fca839d6aff4",
                            "name": "electronic"
                        },
                        {
                            "disambiguation": "",
                            "id": "ffbc9907-c9be-4ace-876b-b7fd5b9d51f9",
                            "count": 4,
                            "name": "industrial rock"
                        }
                    ],
                    "tags": [
                        {
                            "name": "american",
                            "count": 1
                        },
                        {
                            "name": "electronic",
                            "count": 1
                        },
                        {
                            "name": "industrial rock",
                            "count": 4
                        },
                        {
                            "count": 2,
                            "name": "supergroup"
                        }
                    ],
                    "aliases": [],
                    "type": "Group",
                    "id": "11137c88-a9a2-4ffa-a97d-fb058c6d6ce2",
                    "disambiguation": "industrial supergroup",
                    "sort-name": "Pigface",
                    "name": "Pigface"
                }
            }
        ],
```
This is one attribute with multiple values from an instance of an "artist". Since it is a nested list/dictionary, it can be infinitely long. It is impossible to give each single value a column in the CSV file. How do we interpret this?

Yueqiao12Zhang:
My idea is to use an algorithm that keeps track of only the "id" and "name" values recursively and give them columns in the CSV.

Yueqiao12Zhang:
For example:
```
"genres": [
{"id": "1", 
"name": "rnb"
},
{"id": "2",
"name": "jazz"
},
{"id": "3",
"name": "soul"
]
```
Goes to:
```
artist_id,genre_id,,,genre_name,,
001,1,2,3,"rnb","jazz","soul"
```

[...]

PR #47, opened 2024-05-23
Musicbrainz json to csv

Yueqiao12Zhang:
Implemented necessary steps for converting the raw file from MusicBrainz into a CSV file. 
The test case used is in the "labels" section of the database. In these data sets, except for the string attributes, many other attributes are dictionaries or lists. The implementations in this pull request only deal with the string attributes and dictionary attributes that contain "id" and "name" as the keys.
Since lists and dictionaries can be infinitely long and it is difficult to convert them fully into a CSV file, we need to discuss about how to convert these attributes.
The dependencies in pyproject.toml are used for future reconciliation services.


musicbrainz/csv/format_json.py
```diff
@@ -0,0 +1,16 @@
+import json
+
+def format() -> None:
+    """
+    Format the downloaded json file, which is not readable for python json library.
```
ahankinson, line 5: It's in json lines format. https://jsonlines.org

```
with open("test2", "r") as f:
    out = [json.loads(m) for m in f]
```

.gitignore
```diff
@@ -1,3 +1,9 @@
 *.pyc
 */__pycache__
-psqldump
\ No newline at end of file
+psqldump
+*/.DS_Store
```
dchiller, line 4: If you have .DS_Store, you don't need */.DS_Store.

dchiller:
It's not clear to me how this PR solves the referenced issue. Could you either edit that issue or create a new one that merging this PR would solve? From the first sentence of your comments, it seems like you are creating a utility "for converting the raw file from MusicBrainz into a CSV file." Something like that, with some addition information about desired input and output, would be a good issue.

dchiller:
Ok, I started to review but I think the more helpful thing would be a bit of a level set. 

I gather that what you are working on right now is a script to extract some data (related to MusicBrainz) from a json file and transform that data so it can be written to a csv file. 

So, I think the first questions I am interested in are related to what the input to this script is and what the output of this script is (as I say in the previous comment, this probably should go in an issue). If I am reviewing your code, if I am a potential future developer of this code, if I am every going to use this code, I want to know these things:

- what is the input data? how do I get hold of it? If this is only meant to take a very specific json file as input at this time, fine, but I need to know what that file is and where to get it so I can use this utility.
- what is the output like? this is a key part of the issue -- how do you know that you have solved an issue if you haven't clarified what the issue is? I am sure in your head you know what you are trying to get the output to look like, but it's important that this be accessible to others.

Yueqiao12Zhang:
> It's not clear to me how this PR solves the referenced issue. Could you either edit that issue or create a new one that merging this PR would solve? From the first sentence of your comments, it seems like you are creating a utility "for converting the raw file from MusicBrainz into a CSV file." Something like that, with some addition information about desired input and output, would be a good issue.

Ok. Since OpenRefine does not read JSON line format files well, but it can read CSV files and reconcile them easily. That's why we have to convert the raw file to a flattened CSV file first.

However, I still have to figure out how to deal with nested dictionaries and lists. I limited the number of all the nested list of parameters to 3, but I think it's still too much. In the most recent push, I chose to ignore the aliases, which have many repetitions, and relations, which are mostly references to outside websites.

musicbrainz/csv/convert_to_csv.py
```diff
@@ -0,0 +1,101 @@
...
+header = ["id"]
+values = []
+
+with open("test2", "r") as f:
+    json_data = [json.loads(m) for m in f]
+
+# print(json_data)
+
+
+def extract(data, first_level: bool = True, key: str = "", value: dict = {}):
```
ahankinson, line 18: I don't know what editor you're using, but whatever it is, you need to configure it to flag things like this. Unless you know what you're doing, you don't use mutable data structures as default values.

https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil

dchiller:
@Yueqiao12Zhang 

Should you unlink #45? It looks like that will be closed with this PR currently.

musicbrainz/csv/convert_to_csv.py
```diff
@@ -0,0 +1,101 @@
...
+    if key != "":
+        first_level = False
+    if "relations" in key or "aliases" in key:
```
dchiller, line 21: `key` is a string, so you are seeing if these are substrings?

musicbrainz/csv/convert_to_csv.py
```diff
@@ -0,0 +1,101 @@
...
+
+        if isinstance(data, str):
+            data = data.replace(",", "")
+            data = data.replace("\r\n", "")
+
+        if data == None:
+            data = ""
+
+        value[key] = data
+        return
```
dchiller, line 68: This function always returns `None`?

musicbrainz/jsonld/looping-approach/get_ids.py
```diff
@@ -17,7 +17,7 @@ def extract_ids(data, valid_formats, ids=set()):
     return ids
 
 # Load the JSON-LD file
-with open('../../data/initial_compact.jsonld', 'r') as json_file:
+with open("./../../data/initial_compact.jsonld", 'r') as json_file:
```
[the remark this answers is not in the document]
Yueqiao12Zhang: I tried to run the code from Van and there was a small bug here. I fixed it.

musicbrainz/jsonld/looping-approach/ids.csv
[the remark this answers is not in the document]
Yueqiao12Zhang: This is from Van. I think this is auto-generated.

musicbrainz/data/extract.py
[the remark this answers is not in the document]
Yueqiao12Zhang: I don't remember changing this, it might be unintentional.

musicbrainz/data/series.json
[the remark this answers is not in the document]
Yueqiao12Zhang: This is a test file.

musicbrainz/csv/convert_to_csv.py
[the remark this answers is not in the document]
Yueqiao12Zhang: I will add them.

musicbrainz/csv/convert_to_csv.py
Yueqiao12Zhang: Yes.

musicbrainz/csv/convert_to_csv.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+if __name__ == "__main__":
+    extract(json_data)
+
+    with open("out.csv", "w") as out:
```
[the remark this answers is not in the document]
Yueqiao12Zhang: I will try it.

musicbrainz/csv/convert_to_csv.py
Yueqiao12Zhang: Since relations are external links to websites and aliases have a bunch of repeated names, I chose to ignore them when testing.

musicbrainz/csv/convert_to_csv.py
Yueqiao12Zhang: Yes, but I think this can be optimized.

Yueqiao12Zhang: It adds all the data into a global variable. Do recursive returns make it better?

musicbrainz/csv/convert_to_csv.py
dchiller: So you want this to return `True` if `key="here_is_some_alias"?

musicbrainz/csv/test3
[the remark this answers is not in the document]
dchiller: So....should it be in this PR or not?

musicbrainz/data/series.json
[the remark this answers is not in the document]
dchiller: Ok, how will another developer know that?

musicbrainz/data/extract.py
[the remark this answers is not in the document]
dchiller: If you don't know why it changed or didn't change it, shouldn't go in the PR

musicbrainz/csv/convert_to_csv.py
dchiller: I'm not necessarily looking for optimized....I'm just trying to understand what this function does. It is hard for me to understand the way it is written.... more comments maybe? Or some discrete functions?

Yueqiao12Zhang: I will add comments.

musicbrainz/csv/convert_to_csv.py
Yueqiao12Zhang: Yes.

musicbrainz/csv/test3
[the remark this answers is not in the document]
Yueqiao12Zhang: Not now since I did not add anything.

musicbrainz/data/series.json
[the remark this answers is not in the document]
Yueqiao12Zhang: I will document it.

musicbrainz/jsonld/looping-approach/ids.csv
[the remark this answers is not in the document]
Yueqiao12Zhang: I changed because I ran Van's code, but it should not be in this PR. It's my mistake.

musicbrainz/csv/test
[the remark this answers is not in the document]
Yueqiao12Zhang: They are test files for this project, which are snippets copied from MusicBrainz's JSON data dumps. I think they should be committed if other developers want to test the project.

musicbrainz/csv/convert_to_csv.py
```diff
@@ -0,0 +1,101 @@
...
+# Input file can be downloaded in
+# https://data.metabrainz.org/pub/musicbrainz/data/json-dumps/20240522-001002/
+# Output file can be imported to OpenRefine
+import json
+import copy
+
+header = ["id"]
+values = []
+
+with open("test2", "r") as f:
```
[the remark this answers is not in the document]
Yueqiao12Zhang: I will document it.

musicbrainz/csv/convert_to_csv.py
```diff
@@ -12,7 +12,7 @@
 values = []
 
 filename = os.path.dirname(__file__)
-inputpath = os.path.join(filename, "test2")
+inputpath = os.path.join(filename, "data/test2")
```
ahankinson, line 15: If you’re using `os.path.join` you don’t need the slash….

.gitignore
```diff
@@ -1,3 +1,8 @@
 *.pyc
 */__pycache__
-psqldump
\ No newline at end of file
+psqldump
+.DS_Store
+cantusdb/*
```
dchiller, line 5:
```suggestion
```

musicbrainz/csv/convert_to_csv.py
```diff
@@ -0,0 +1,131 @@
+# This file converts a JSON Lines file from MusicBrainz to a CSV file
```
dchiller, line 1: This probably should be a docstring (see [here](https://peps.python.org/pep-0257/))

musicbrainz/jsonld/looping-approach/get_ids.py
dchiller, line 1: If this was a bug-fix, this should be documented in the PR description.

musicbrainz/jsonld/looping-approach/ids.csv
dchiller, line 1: Why is this file changed? It is not clear based on documentation in the PR description.

musicbrainz/csv/reconcile.py
dchiller, line 1: This file is reconciling against a wikidata API? How is that relevant to the linked issue or to the PR description?

musicbrainz/jsonld/looping-approach/ids.csv
Yueqiao12Zhang: All accidentally changed files are deleted.

musicbrainz/csv/convert_to_csv.py
Yueqiao12Zhang: A docstring for the extract function?

musicbrainz/data/extract.py
dchiller, line 1: Do you really want to delete this file?

musicbrainz/jsonld/looping-approach/get_ids.py
dchiller, line 1: Same with this...do you really want to delete?

musicbrainz/jsonld/looping-approach/ids.csv
dchiller, line 1: And with this...deleting?

musicbrainz/csv/reconcile.py
dchiller, line 1: You said this should be a new issue, but the file is still here?

musicbrainz/csv/data/test
dchiller, line 1: Do `test` and `test2` not have extensions?

musicbrainz/csv/data/out-csv.csv
dchiller, line 1: `out.csv` is ignored but `out-csv.csv` is not?

musicbrainz/csv/convert_to_csv.py
dchiller: Or for the module... so you could put the docstring for the module as the first thing in the file. Since you have the comments here, that would be the most similar thing.

musicbrainz/csv/convert_to_csv.py
```diff
@@ -0,0 +1,131 @@
...
+        data : can be anytype, parsed based on its type
+        value : records the informations of the current dictionary
+        first_level : records if the current level is the first level of the JSON file.
+        key : the current key that will be added to a dictionary
+    '''
+    if key != "":
+        first_level = False
+        
+    if "relations" in key or "aliases" in key or "tags" in key:
+        # ignore relations and aliases to make output simplier
```
dchiller, line 37:
```suggestion
        # ignore relations, aliases, and tags to make output simplier
```

musicbrainz/csv/convert_to_csv.py
```diff
@@ -0,0 +1,131 @@
...
+        global header
+        if key not in header:
+            header.append(key)
+
+        v = data
+        if isinstance(data, str):
+            v = v.replace(",", "_")
+            v = v.replace("\r\n", "")
+
+        if data is None:
```
dchiller, line 100: Could this be an `elif`?

musicbrainz/data/extract.py
Yueqiao12Zhang: No, I just want to delete this file on this branch.

musicbrainz/jsonld/looping-approach/get_ids.py
Yueqiao12Zhang: Same here.

musicbrainz/csv/reconcile.py
Yueqiao12Zhang: Where should I put it?

musicbrainz/csv/data/out-csv.csv
Yueqiao12Zhang: They are renamed. The out-csv.csv is reconciled in OpenRefine by hand.

musicbrainz/csv/convert_to_csv.py
```diff
@@ -0,0 +1,131 @@
...
+    extract(json_data, {})
+
+    with open(outputpath, "w") as out:
+        # write header
+        writer = csv.writer(out)
+        
+        header.sort(key=len)
+        # id must be the first column of the csv file. Reorder the header.
+        header.remove("id")
+        header.insert(0, "id")
```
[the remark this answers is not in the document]
Yueqiao12Zhang: Should I sort by length? Or is sort in alphabetical order better?

musicbrainz/csv/convert_to_csv.py
```diff
@@ -0,0 +1,131 @@
...
+
+    else:
+        # if data is not a collection, we parse it and add to the current value dictionary.
+        global header
+        if key not in header:
+            header.append(key)
+
+        v = data
+        if isinstance(data, str):
+            v = v.replace(",", "_")
```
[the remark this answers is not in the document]
Yueqiao12Zhang: They are, I will fix it.

musicbrainz/csv/convert_to_csv.py
Yueqiao12Zhang: This clause is just preventing the program from adding a None value to the dictionary. If it's not None, then we add the original value to the dictionary.

musicbrainz/csv/convert_to_csv.py
```diff
@@ -1,69 +1,134 @@
...
+        if data is None:
+            v = ""
+
+        try: 
+            value[key]
+        except KeyError:
+            value[key] = v
+        else:
+            if isinstance(value[key], list): (value[key]).append(v)
+            else: value[key] = [value[key]] + [v]
```
ahankinson, line 98: This is really hard to debug, and it's absolutely bonkers to me that that you would raise an exception just to get the functionality of an `if` statement, but I'm pretty sure it can be rewritten something like this:

```suggestion
    if key in value:
            if isinstance(value[key], list): 
                value[key].append(v)
            else: 
                value[key] = [value[key]] + [v]
    else:
        value[key] = v
```

I won't comment on how the code is actually functioning...

musicbrainz/csv/convert_to_csv.py
```diff
@@ -105,6 +112,16 @@ def extract(data, value: dict, first_level: bool = True, key: str = ""):
         return
     
 def convert_dict_to_csv(dictionary_list, filename):
```
ahankinson, line 114:
```suggestion
def convert_dict_to_csv(dictionary_list: list, filename: str) -> None:
```

musicbrainz/csv/convert_to_csv.py
```diff
@@ -105,6 +112,16 @@ def extract(data, value: dict, first_level: bool = True, key: str = ""):
...
+    """
+    (list, str) -> None
```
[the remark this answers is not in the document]
ahankinson: Also, just a matter of "thinking about code":  a function that returns `None` is a red flag. That means that it's got "side effects" and is modifying data outside of its inner scope. For classes this is kinda expected, since the methods on a class can interact with class and instance variables. But for standalone functions, this indicates that your function is changing something somewhere else. 

This, of course, is one of the main source of bugs in code: Variables being modified in many places, and no "ownership" of the data at any given point in time. So you may find your data being in an unexpected state because the application of the processing on that data may not be in an order that you expect it in. 

The TL;DR is that, unless you're writing code, your functions should always return a value, and given the same arguments, the return value should be the same. This will help you avoid bugs you didn't even know you could have!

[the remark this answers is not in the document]
Yueqiao12Zhang: It is not modifying outside variables, it is a serialization process of writing data into a file.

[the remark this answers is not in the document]
ahankinson: If it’s returning None then that is not possible. You are modifying a single instance of a dict which is declared outside of the scope of this function.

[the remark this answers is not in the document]
Yueqiao12Zhang: I am only reading from that dict, how am I modifying it?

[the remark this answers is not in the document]
ahankinson: If the return value of your function is None, the. It must be modifying data somewhere, right? Otherwise there is no effect of your function….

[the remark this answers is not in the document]
ahankinson: Sorry, I just noticed that this was attached to the wrong function. I was thinking about the `extract` method above, but I notice that my comment is on this function. (It's the same observation, just on a different method).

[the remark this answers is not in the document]
Yueqiao12Zhang: Yes, the extract function is modifying a dictionary defined outside.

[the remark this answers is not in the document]
ahankinson: You can't. Not easily, anyway. In a purely functional language, you would define your recursion to return the dictionary at each step, which would then get modified in the next step, and returned, until it ended. That's not how you have designed yours. 

What I haven't mentioned (up to this point) is that using the `global` keyword is generally frowned upon in Python, for exactly this same reason. It makes for code that is difficult to debug, because the value can be changed in many places. 

No action needed for this (you've had enough changes on this file) but when you write your next converter, try not to use `global`...

Yueqiao12Zhang:
All clean-ups are done.

musicbrainz/csv/convert_to_csv.py
```diff
@@ -1,69 +1,161 @@
...
+    '''
+    (data, dict, bool, str) -> None
     
-for element in json_data:
-    value = {}
-    for k in element:
-        v = element[k]
+    Extract info from JSON Lines file and add a finite number of them into a list of dictionaries.
+    Arguments:
+        data : can be any type, parsed based on its type
```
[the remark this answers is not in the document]
Yueqiao12Zhang: Not tuple.

musicbrainz/csv/convert_to_csv.py
```diff
@@ -1,69 +1,161 @@
...
-cnt = 0
 values = []
 
-format_json.format()
-    
-with open("json_init.json", 'r') as json_file:
-    json_data = json.load(json_file)
-    # print(json_data)
+DIRNAME = os.path.dirname(__file__)
+inputpath = os.path.join(DIRNAME, "data", "test2")
```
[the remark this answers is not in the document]
Yueqiao12Zhang: Now the input depends on the arguments.

musicbrainz/csv/convert_to_csv.py
```diff
@@ -1,69 +1,161 @@
+# This file converts a JSON Lines file from MusicBrainz to a CSV file
+# For OpenRefine Reconciling
+# Input file can be downloaded in
+# https://data.metabrainz.org/pub/musicbrainz/data/json-dumps/20240522-001002/
```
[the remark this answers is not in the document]
Yueqiao12Zhang: It's fixed.

dchiller (approved):
Ok, approving now. 

@Yueqiao12Zhang and I had a good discussion about this this morning, including directions for future PR's.

PR #57, opened 2024-06-10
Csv2rdf single subject bug fix and generalization and move files

Yueqiao12Zhang:
This script now uses the correct mapping filename generated by another script.
Adds the .history directory (used by the Local History VSCode extension) to .gitignore


.gitignore
```diff
@@ -3,3 +3,4 @@
 psqldump
 .DS_Store
 cantusdb/turtle/out_rdf.ttl
+.history/*
```
dchiller, line 6: This should be removed from the PR. There is no explanation of why it is here and is not referenced by any other file in the PR.

cantusdb/turtle/csv2rdf_single_subject.py
dchiller, line 1: This looks very similar to `csv2rdf/csv2rdf_single_subject.py`... why should the repo have both of these files?

csv2rdf/csv2rdf_single_subject.py
```diff
@@ -0,0 +1,75 @@
...
+                        obj = Literal(element)
+
+                    g.add((key_attribute, predicates[i], obj))
+
+    return g
+
+# Run this file in the command line.
+# Args = the input reconciled csv files using openrefine
+# Convert the CSV data to Turtle format
+# out_rdf.ttl can be safely imported into Virtuoso.
```
dchiller, line 68: This would probably be better as a module docstring.

csv2rdf/csv2rdf_single_subject.py
```diff
@@ -0,0 +1,75 @@
+import csv
+import validators
+import sys
+import json
+import os
+from rdflib import Graph, URIRef, Literal
```
dchiller, line 6: This script requires a third-party library. Is that documented either in a Readme, a requirements.txt or a pyproject.toml file?

csv2rdf/csv2rdf_single_subject.py
```diff
@@ -0,0 +1,75 @@
...
+from rdflib.namespace import RDF
+from typing import Optional, List
+
+DIRNAME = os.path.dirname(__file__)
+mapping_filename = os.path.join(DIRNAME, 'mapping.json')
+dest_filename = os.path.join(DIRNAME, 'out_rdf.ttl')
```
dchiller, line 12: `mapping.json` does not correspond to any of the other files in this PR, which makes me wonder again whether these two wouldn't be better as command-line arguments.

csv2rdf/csv2rdf_single_subject.py
```diff
@@ -0,0 +1,75 @@
...
+            csv_reader = csv.reader(csv_file)
+
+            header = next(csv_reader)
+            header_without_subject = header[1:]
+            predicates = []
+            for column in header_without_subject:
+                if column in ontology_dict:
+                    predicates.append(URIRef(ontology_dict[column]))
+
+            try:
```
dchiller, line 38: This should be a call to the dictionary's `get` method.

cantusdb/turtle/csv2rdf_single_subject.py
Yueqiao12Zhang: There should not be both files, I think it's a mistake in the rebasing process. I will remove it.

csv2rdf/csv2rdf_single_subject.py
Yueqiao12Zhang: Ok.

csv2rdf/csv2rdf_single_subject.py
Yueqiao12Zhang: It's in pyproject.toml.

csv2rdf/csv2rdf_single_subject.py
Yueqiao12Zhang: It's also a mistake in the rebasing process. I missed one commit.

csv2rdf/csv2rdf_single_subject.py
Yueqiao12Zhang: I changed the logic here.

.gitignore
dchiller: That's fine. Can you add to the description of this PR something like: "Adds the .history directory (used by the ______ VSCode extension to .gitignore"

csv2rdf/csv2rdf_single_subject.py
dchiller: Then the `pyproject.toml` should be in this PR.

.gitignore
Yueqiao12Zhang: This is changed in other PRs as well. I will also document in other PRs.

csv2rdf/csv2rdf_single_subject.py
Yueqiao12Zhang: The pyproject.toml is in the main branch. It's not changed.

Yueqiao12Zhang: These lines are moved directly from another file, they should also be the same in the main branch, except they are in the original location.

csv2rdf/csv2rdf_single_subject.py
```diff
@@ -1,3 +1,7 @@
+# Run this file in the command line.
+# Args = the input reconciled csv files using openrefine
+# Convert the CSV data to Turtle format
+# out_rdf.ttl can be safely imported into Virtuoso.
```
dchiller, line 4:
```suggestion
"""
Run this file in the command line to convert csv files containing
reconciled data to turtle files.
The script takes an arbitrary number of positional arguments.
The first argument is a path to a relation mapping file for the
data in the csvs. The relations mapping file is a json file where the
keys are headers in the csv file and the values are URIs to Wikidata properties.
The following arguments are paths to reconciled csv files to be converted.
The script creates a file containing the turtle in the directory containing this
file called 'out_rdf.ttl'.
"""
```
Will you check if this seems correct. I made it a docstring, added more details, and corrected the description of the command line arguments.

csv2rdf/csv2rdf_single_subject.py
```diff
@@ -23,8 +30,13 @@ def convert_csv_to_turtle(filenames: List[str]) -> Graph:
     g = Graph()
 
     ontology_dict = json.load(open(mapping_filename, "r"))
+    try:
```
dchiller, line 33: this is still not `get`?

csv2rdf/csv2rdf_single_subject.py
Yueqiao12Zhang: All is good except that the values are URIs to either Wikidata, Schema.org, or other used properties from the database websites (Wikidata is the better one).

dchiller: Ok, that makes sense. Can you update with that?

Yueqiao12Zhang: Yes, it's updated

csv2rdf/csv2rdf_single_subject.py
```diff
@@ -23,8 +37,10 @@ def convert_csv_to_turtle(filenames: List[str]) -> Graph:
     g = Graph()
 
     ontology_dict = json.load(open(mapping_filename, "r"))
+    ontology_list = ontology_dict.get("entity_type")
 
-    for filename in filenames:
+    for i, filename in enumerate(filenames):
+        ontology_type = ontology_list[i]
```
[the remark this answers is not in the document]
Yueqiao12Zhang: If in the JSON file, the 'ontology_list' can be empty, but it will always be a list, then will it be None?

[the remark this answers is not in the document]
dchiller: In line 40, you have `ontology_list = ontology_dict.get("entity_type")`. If `ontology_dict` has no key `entity_type` then, `ontology_list` will be `None`.

csv2rdf/csv2rdf_single_subject.py
Yueqiao12Zhang: You are right, I confused with the other pyproject.toml.

csv2rdf/csv2rdf_single_subject.py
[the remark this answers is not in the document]
Yueqiao12Zhang: If we use the new get_relations.py from the other PR, then we guarantee to have a list of "entity_type".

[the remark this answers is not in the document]
dchiller: Add a comment here then, please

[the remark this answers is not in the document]
Yueqiao12Zhang: Ok, but if we have 3 input filenames and 2 entity_types, then there might be a problem.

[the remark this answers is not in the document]
Yueqiao12Zhang: I will try to fix this.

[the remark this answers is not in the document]
dchiller: Ok, I think this might be something to come back to if we turn this into an automated pipeline, but I'm good for now. I will approve.

PR #64, opened 2024-06-11
Generate mapping file

Yueqiao12Zhang:
This PR creates a new script that generates an empty mapping file with all the necessary keys. 
Since all headers are used as predicates in the RDF triples, they need to be reconciled to some other URIs. This mapping file will contain a dictionary with all header cells as keys, and all values are empty. Users should reconcile them according to Wikidata / schema.org.


csv2rdf/get_relations.py
```diff
@@ -0,0 +1,22 @@
+import csv
```
dchiller, line 1: This needs a docstring to explain usage...

csv2rdf/get_relations.py
```diff
@@ -0,0 +1,22 @@
...
+    with open(filename, "r") as csv_file:
+        csv_reader = csv.reader(csv_file)
+        header = next(csv_reader)
+        
+    header.append("entity_type")
+
+    for item in header:
+        dt[item] = ""
+        
+    dt["entity_type"] = []
```
dchiller, line 19: You don't need to do this with every loop.

dchiller:
@Yueqiao12Zhang See comments above. I have also been lax about commit messages thus far, but going forward please try to be more descriptive. See, for example, the document that Ich sent out about commit messages earlier this week. If you have multiple commits in a row with the same message, that is probably a sign that your messages need to be more specific!

Yueqiao12Zhang:
> @Yueqiao12Zhang See comments above. I have also been lax about commit messages thus far, but going forward please try to be more descriptive. See, for example, the document that Ich sent out about commit messages earlier this week. If you have multiple commits in a row with the same message, that is probably a sign that your messages need to be more specific!
> 
> If you were to solve my two comments in separate commits, they could be something like: `docs: add module docstring to csv2rdf/get_relations.py` and `refactor: stop setting entity_type for every file in csv2rdf/get_relations.py`

What if I do both things in one commit?

dchiller:
> What if I do both things in one commit?

Then write a commit message that describes both? All I'm saying is that "Update [filename]" is never a good commit message because it is too broad. There's always something more specific.

csv2rdf/get_relations.py
```diff
@@ -0,0 +1,30 @@
...
+output_name = "mapping.json"
+dt = {}
+dt["entity_type"] = []
+
+for filename in filenames:
+    with open(os.path.abspath(filename), "r") as csv_file:
+        csv_reader = csv.reader(csv_file)
+        header = next(csv_reader)
+        
+    header.append("entity_type")
```
dchiller, line 24: You also don't need to do this in the loop?

dchiller:
Are you using `black` and `pylint`? There are multiple things in this file that those are flagging for me!

csv2rdf/get_relations.py
```diff
@@ -0,0 +1,30 @@
...
+import sys
+import os
+
+filenames = sys.argv[1:]
+output_name = "mapping.json"
+dt = {}
+dt["entity_type"] = []
+
+for filename in filenames:
+    with open(os.path.abspath(filename), "r") as csv_file:
```
dchiller, line 20: Why do you need the abspath call here?

Yueqiao12Zhang: If we move the test files elsewhere (a test folder inside this directory) and do not use the abspath call, then Python is not able to find the file.

dchiller (approved):

PR #78, opened 2024-07-03
Simssa csv

Yueqiao12Zhang:
Use the previous CSV to convert Simssa DB to RDF.


csv2rdf/out_rdf.ttl
dchiller, line 1: Why would this file have changed?

csv2rdf/simssa/mapping.json
dchiller, line 1: I have some questions about the choice for some of these...for example, `sacred_or_secular` or `genre_style`. But since this is just testing, does it matter?

simssadb/csv/data_detail.md
dchiller, line 1: How did you do this separation into separate csv files? Manually?

csv2rdf/out_rdf.ttl
Yueqiao12Zhang: In the csv2rdf script, the output is named out_rdf.ttl, regardless of the input. So the original out_rdf.ttl is for another database.

csv2rdf/simssa/mapping.json
Yueqiao12Zhang: I think it does matter. I am a little incautious when I reconcile things, I will be more careful after.

simssadb/csv/data_detail.md
Yueqiao12Zhang: Yes.

csv2rdf/simssa/mapping.json
dchiller: Ok, I think that makes sense to me. :)

simssadb/csv/data_detail.md
dchiller: Ok, could you just add the clarification that you did this separation manually in this document?

csv2rdf/out_rdf.ttl
dchiller: So then this was overwriting the results from a different database? If that is the case, then I think this file should be added to `.gitignore`

dchiller (approved):

PR #80, opened 2024-07-08
Musicbrainz json to csv fix

Yueqiao12Zhang:
Based on the inspection issue by Junjun, I adapted some changes to the musicbrainz json to csv. MusicBrainz csv to rdf will be changed in another PR.


musicbrainz/csv/convert_to_csv.py
```diff
@@ -90,9 +89,11 @@ def extract(data, value: dict, first_level: bool = True, key: str = ""):
                     # extract its id
                     keywords = key.split("_")
                     word = keywords[-1]
-                    id = data["id"]
+                    if word.endswith("s"):
```
ahankinson, line 92: Really?  “Beatles” -> “Beatle”?

musicbrainz/csv/convert_to_csv.py
```diff
@@ -120,13 +130,13 @@ def extract(data, value: dict, first_level: bool = True, key: str = ""):
...
     else:
         # if data is not a collection, we parse it and add to the current value dictionary.
-        global header
         if key not in header:
             header.append(key)
 
         v = data
         if isinstance(data, str):
-            v = v.replace(",", "\,")
+            if "," in v:
```
ahankinson, line 138: You’re working way too hard. The Python CSV writer auto-escapes values with commas.

musicbrainz/get_genre.py
```diff
@@ -0,0 +1,31 @@
...
+import time
+import requests
+import pandas as pd
+
+URL = "https://musicbrainz.org/ws/2/genre/all"
+PARAMS = dict(fmt="json", limit="50", offset="0")
+
+data = []
+
+for i in range(0, 1947, 50):
```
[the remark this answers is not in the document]
dchiller: I'm seeing this in the api response: 

image

musicbrainz/get_genre.py
```diff
@@ -0,0 +1,31 @@
+"""
+Script for retrieving the "genre" dumps since they are not available online.
+"""
+import time
+import requests
+import pandas as pd
+
+URL = "https://musicbrainz.org/ws/2/genre/all"
+PARAMS = dict(fmt="json", limit="50", offset="0")
```
dchiller, line 9:
```suggestion
PARAMS = dict(fmt="json", limit="50")
```

musicbrainz/get_genre.py
```diff
@@ -0,0 +1,31 @@
...
+data = []
+
+for i in range(0, 1947, 50):
+    PARAMS["offset"] = i
+    resp = requests.get(url=URL, params=PARAMS, timeout=100)
+    temp = resp.json()
+
+    try:
+        data += dict(temp)["genres"]
+    except KeyError:
```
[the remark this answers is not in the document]
Yueqiao12Zhang: Sometimes, the request returns an error. This is to see the error message and not interrupting the requests.

musicbrainz/get_genre.py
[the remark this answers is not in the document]
Yueqiao12Zhang: Yes, I will change this to requesting the "genre-count".

musicbrainz/get_genre.py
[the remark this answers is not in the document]
dchiller: Right, but what error does it return. Is it an error that we should be mitigating in some way?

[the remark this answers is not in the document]
Yueqiao12Zhang: It returns a dictionary, the same format as a normal response, but the only key is "error", not "genre".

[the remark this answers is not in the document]
dchiller: Why does it do that? When it does that do we just not get those genres? 

What I am getting at is that if there is a predictable error that is occurring, then it makes sense to have something in the code that mitigates this error or helps one know what to do in case of that error.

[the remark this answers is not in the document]
Yueqiao12Zhang: > Why does it do that? When it does that do we just not get those genres?

I don't really get the pattern. Sometimes it happens and sometimes it does not. My way in this case is just avoid the interruption that it might cause.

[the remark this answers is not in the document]
dchiller: Is it some sort of timeout? 

What do you do with that set of genres? Do you just not include them?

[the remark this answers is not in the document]
Yueqiao12Zhang: I think so. If error message appears, then I just run the script again to try again.

Yueqiao12Zhang:
Also, there is a Wikidata link on the webpage of a genre, but there is no Wikidata link in the genres response. Where is the reconciled Wikidata link stored?

musicbrainz/get_genre.py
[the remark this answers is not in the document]
Yueqiao12Zhang: This is the dictionary that error occurred.
`{'error': 'Your requests are exceeding the allowable rate limit. Please see http://wiki.musicbrainz.org/XMLWebService for more information.'}`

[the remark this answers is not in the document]
dchiller: Yeah, ok...so the error happens because you are requesting more frequently than the rate limit they have imposed...so you could:
- request less frequently
- when handling an exception, slow down the rate of your requests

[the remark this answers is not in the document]
ahankinson: - check the status code of the HTTP requests -- that will let you know if the request succeeded or not (if it was `200` then yes; anything else, no.)
 - If you do receive a [`503`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503) you can catch that and re-try the request after waiting. 
 - [set a friendly `User-Agent` header](https://wiki.musicbrainz.org/MusicBrainz_API/Rate_Limiting#How_can_I_be_a_good_citizen_and_be_smart_about_using_the_Web_Service?)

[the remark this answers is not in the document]
dchiller: See here: https://requests.readthedocs.io/en/latest/user/quickstart/#custom-headers

musicbrainz/csv/get_genre.py
```diff
@@ -0,0 +1,47 @@
...
+max_records = dict(requests.get(url=URL, params=PARAMS, timeout=500).json())[
+    "genre-count"
+]
+
+for i in range(0, max_records, 50):
+    PARAMS["offset"] = i
+    resp = requests.get(url=URL, params=PARAMS, timeout=500)
+    if resp.status_code == 503:
+        time.sleep(1)
+        resp = requests.get(url=URL, params=PARAMS, timeout=500)
```
dchiller, line 23: What if this fails a second time?

musicbrainz/csv/get_genre.py
```diff
@@ -0,0 +1,47 @@
...
+
+    temp = resp.json()
+    data += dict(temp)["genres"]
```
dchiller, line 26: I don't think you need `dict` here

musicbrainz/csv/get_genre.py
```diff
@@ -18,14 +18,13 @@
 for i in range(0, max_records, 50):
     PARAMS["offset"] = i
     resp = requests.get(url=URL, params=PARAMS, timeout=500)
-    temp = resp.json()
+    if resp.status_code == 503:
```
ahankinson, line 21: So you’re only slowing down your requests once you hit the rate limit, despite knowing that you’re going to be rate limited?

Why not slow down all your requests to something that aligns with their documentation? 1/50th of a second if you want to be on the edge, 1/25th if you want a bit of room.

musicbrainz/csv/get_genre.py
```diff
@@ -18,14 +18,13 @@
...
+        time.sleep(1)
+        resp = requests.get(url=URL, params=PARAMS, timeout=500)
```
ahankinson, line 23: What happens if this one doesn’t succeed?

musicbrainz/get_genre.py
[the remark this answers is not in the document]
ahankinson: I know how your code works. I’m saying you’re going to be spamming them until you get rate limited, then you’re going to get hard limited.

why not follow their documentation and keep the number of all your requests to under 50/sec?

musicbrainz/csv/get_genre.py
```diff
@@ -0,0 +1,47 @@
...
+df = pd.DataFrame(data)
+df = df[["id", "name", "disambiguation"]]
+
+df["id"] = "https://musicbrainz.org/genre/" + df["id"].astype(str)
+df.rename(columns={"id": "genre_id"}, inplace=True)
+relations_wiki = []
+
+for rec in df["genre_id"]:
+    resp_wiki = requests.get(rec, timeout=500)
+    resp_wiki.raise_for_status()
```
dchiller, line 38: How are you using this? You aren't handling an exception...

musicbrainz/csv/get_genre.py
Yueqiao12Zhang: I will experiment

musicbrainz/csv/get_genre.py
```diff
@@ -0,0 +1,62 @@
+"""
+Script for retrieving the "genre" dumps since they are not available online.
+"""
+
+import time
+import requests
+from requests.exceptions import HTTPError
+import pandas as pd
+from bs4 import BeautifulSoup
```
dchiller, line 9: This dependency needs to be documented.

musicbrainz/csv/get_genre.py
```diff
@@ -0,0 +1,62 @@
...
+
+URL = "https://musicbrainz.org/ws/2/genre/all"
+PARAMS = dict(fmt="json", limit="50")
+
+data = []
+max_records = dict(requests.get(url=URL, params=PARAMS, timeout=500).json())[
```
dchiller, line 15: Do you need dict here? The json method will return a python data structure (dict or list) depending on the json object returned.

musicbrainz/csv/get_genre.py
```diff
@@ -0,0 +1,62 @@
...
+    "genre-count"
+]
+
+for i in range(0, max_records):
+    PARAMS["offset"] = i
```
dchiller, line 20: This looks like your `offset` parameter is incrementing by 1 but your `limit` is 50...won't you request the same genre over and over again?

musicbrainz/csv/get_genre.py
```diff
@@ -0,0 +1,62 @@
...
+    for j in range(0, 3):
```
dchiller, line 21: Why did you choose 3?

fujinaga:
@Yueqiao12Zhang "3" is a magic number!

musicbrainz/csv/get_genre.py
Yueqiao12Zhang: If I don't give it a number for how many times I want to retry a request, then how should I deal with errors?

musicbrainz/csv/get_genre.py
Yueqiao12Zhang: If I my offset increase 50 each time and my limit is 50, then is it making 50 API calls or is it making 1 API call?

musicbrainz/csv/get_genre.py
Yueqiao12Zhang: In the pyproject.toml?

musicbrainz/csv/get_genre.py
dchiller: 1 request.get call = 1 API call?

I'm not sure I understand your question.

Yueqiao12Zhang: 1 request.get API call that requests 50 record at once
versus
1 request.get API call that only requests 1 record

musicbrainz/csv/get_genre.py
dchiller: I didn't say don't have a certain number of times to retry...

Walk me through in words what happens here if the request returns a 503 error.

musicbrainz/csv/get_genre.py
dchiller: OK... my only question is is this duplicating the genres you are getting back?

musicbrainz/csv/get_genre.py
Yueqiao12Zhang: Ok. Let's say we request 50 records per iteration. The counter i increment by 50 each loop. 
In the nested loop, we try to make a request.get call. If it does not raise_for_status, then the nested loop breaks and continues. 
If it raise an error, it goes into the except clause. If the error is 503, then we sleep for j seconds, where j is the number of requests.get API calls we already made to the same set of 50 records, and we try again. Once we get 3 errors continuously for one set of 50 records, we raise an error.

Yueqiao12Zhang:
> @Yueqiao12Zhang "3" is a magic number!

It is the maximum number of retries of API calls we want to make to one set of genres data.

musicbrainz/csv/get_genre.py
dchiller: It looks to me that after the 3rd `503` error you will exit the loop without raising an error, no?

Yueqiao12Zhang: Yes.

dchiller:
> It is the maximum number of retries of API calls we want to make to one set of genres data.

Yes! But the problem with magic numbers is not that you might have a number constant (in this case, it might be that you only want to try this 3 times...that's fine), it is that I or anyone else looking at or debugging the code has any idea where the 3 came from! 

One best practice would be to create a constant variable. For example:
```python
MAX_REQUEST_RETRIES = 3
for j in range(MAX_REQUEST_RETRIES):
    ...
```

musicbrainz/csv/get_genre.py
Yueqiao12Zhang: Right now it works fine, I changed i back to jumping in 50s.

musicbrainz/csv/get_genre.py
ahankinson: The point of rate limiting your requests is never to actually get a 503! If you get a 503 you’re abusing the service! You need to write your code so that you do it as fast as you can but no faster.

Yueqiao12Zhang: Right now it goes in the rate of 1/25 secs per request. If it get a 503, it waits for the number of seconds that is the number of times it reties.

dchiller:
@Yueqiao12Zhang Please separate out the `get_genre.py` file and related files into a different PR with a relevant issue attached. Then, create an issue that detail whatever "fixes" were needed in the Musicbrainz json to csv process so I can review that part. Thanks!

Yueqiao12Zhang:
> @Yueqiao12Zhang Please separate out the `get_genre.py` file and related files into a different PR with a relevant issue attached. Then, create an issue that detail whatever "fixes" were needed in the Musicbrainz json to csv process so I can review that part. Thanks!

The problems need to be fixed are in the linked issue in this PR by Junjun. The genre.csv is the generated file from get_genre.py, but since it is processed like a test file, it's updates are linked to the mappings and other test files. Therefore I left it there are you can treat it as other test files.

csv2rdf/thesession/out_rdf.ttl
dchiller, line 1: There's no reference in the issue or the PR description to the session. Make a new issue for this.

csv2rdf/simssa/mapping.json
dchiller, line 1: There's no reference in the issue or PR to simssa. Make a new issue for this.

musicbrainz/csv/convert_to_csv.py
```diff
@@ -28,13 +28,13 @@
 
 entity_type = sys.argv[2]
 inputpath = os.path.join(DIRNAME, sys.argv[1])
-outputpath = os.path.join(DIRNAME, "data", f"{entity_type}.csv")
+outputpath = os.path.join(DIRNAME, "../data/output", f"{entity_type}.csv")
```
ahankinson, line 31: Why are you using slash separators within an `os.path.join` call? the whole point of `os.path.join` is that it adds OS-appropriate path separators for you!

dchiller (approved):

Issue #79, opened 2024-07-04
MusicBrainz RDF needs some improvements


candlecao:
2. "artist-credit_artist_genres_id" is not embodied at the front end, removing "artist-credit_artist_genres_id" column in recording.csv. Since for each entity type, we can track it to other CSVs using it's ID, so there's no need to extract further. Such as artist_genre_id in recording's CSV, we don't need this column since we have an genre_id for artist in artist's CSV. I inspected all CSVs and there is no special case for exception. 

3. Fixing many incorrectly reconciled cells. 
eg., https://musicbrainz.org/genres/eacfa027-2fad-413f-a2f1-80fa43674f0b, https://musicbrainz.org/genres/cc28f5aa-5e19-487c-88e1-e56b022f1fdd, https://musicbrainz.org/genres/0e3fc579-2d24-4f20-9dae-736e1ec78798)

They all need to perform the change: /genres/ -> /genre/. They will be fixed in the conversion script.

4. the rdf:type of "recording" https://www.wikidata.org/wiki/Q155171. Recording is changed to https://www.wikidata.org/wiki/Q15975575.

5. the rdf:type of "area" https://www.wikidata.org/wiki/Q11500. Area is changed to https://www.wikidata.org/wiki/Q2221906.

6. Property of area is changed to https://www.wikidata.org/wiki/Property:P276.

7. Relations_wiki property change: P1687 -> P2888 (exact match) of Wikidata. Better matching the virtuoso and reasoning machine pattern.

8. In Wikidata and SPARQL, as to property, the standard format is http://www.wikidata.org/prop/direct/Pxxx, instead of https://www.wikidata.org/wiki/Property:Pxxx.

In a similar way, as to entity, the standard format is http://www.wikidata.org/entity/Qxxx, instead of https://www.wikidata.org/wiki/Qxxx. 

For these format changes, they also apply to all other databases' mappings.

9. Use music genre Q188451 rather than genre Q483394 for https://musicbrainz.org/doc/Genre.

10. Re-structure the repository:
- add urllib3 and bs4 to the dependency management documentation and remove the poetry.lock and pyproject.toml inside the Musicbrainz repo.
- move the test files, scripts, and docs around to match a standard pattern of other database repositories.

11. Output RDF subjects should be specific about their types when they are not URIs or Literals, such as Integers and Booleans.

candlecao:
[...]

PR #111, opened 2024-07-29
Musicbrain getgenre

Yueqiao12Zhang:
Adding a get_genre.py to fetch the genres from MusicBrainz.


dchiller:
@Yueqiao12Zhang why did you link #79?

musicbrainz/csv/get_genre.py
```diff
@@ -0,0 +1,64 @@
+"""
+Script for retrieving the "genre" dumps since they are not available online.
```
dchiller, line 2:
```suggestion
Script for retrieving the "genre" dumps.
```
They are available online...that's how you are getting them in this script.

musicbrainz/csv/get_genre.py
```diff
@@ -0,0 +1,64 @@
...
+from requests.exceptions import HTTPError
+import pandas as pd
+from bs4 import BeautifulSoup
+
+URL = "https://musicbrainz.org/ws/2/genre/all"
+PARAMS = dict(fmt="json", limit="50")
+MAX_REQUEST_RETRIES = 3
+
+data = []
+max_records = requests.get(url=URL, params=PARAMS, timeout=500).json()["genre-count"]
```
dchiller, line 16: Timeout very long! (more than 5 minutes!)

musicbrainz/csv/get_genre.py
```diff
@@ -0,0 +1,64 @@
...
+
+for i in range(0, max_records, 50):
+    PARAMS["offset"] = i
+    for j in range(MAX_REQUEST_RETRIES):
+        try:
+            resp = requests.get(url=URL, params=PARAMS, timeout=500)
```
dchiller, line 22: Timeout very long!

musicbrainz/csv/get_genre.py
```diff
@@ -0,0 +1,64 @@
...
+URL = "https://musicbrainz.org/ws/2/genre/all"
+PARAMS = dict(fmt="json", limit="50")
+MAX_REQUEST_RETRIES = 3
+
+data = []
+max_records = requests.get(url=URL, params=PARAMS, timeout=500).json()["genre-count"]
+
+for i in range(0, max_records, 50):
+    PARAMS["offset"] = i
+    for j in range(MAX_REQUEST_RETRIES):
```
dchiller, line 20: We were having this discussion in the other thread, I think, but I don't think the result of that made its way here yet. 

Reviewing the (API documentation for MusicBrainz)[https://musicbrainz.org/doc/MusicBrainz_API/Rate_Limiting]:
1. They want an informative `User-Agent` header so they can contact us if something is going wrong. 
2. Read the information about rate limiting based on source IP. Have we constructed this to be within those limits?
3. If they reach the global limit, how does this script respond?

musicbrainz/csv/get_genre.py
```diff
@@ -0,0 +1,64 @@
...
+
+df = pd.DataFrame(data)
+df = df[["id", "name"]]
+
+df["id"] = "https://musicbrainz.org/genre/" + df["id"].astype(str)
+df.rename(columns={"id": "genre_id"}, inplace=True)
+relations_wiki = []
+
+for rec in df["genre_id"]:
+    resp_wiki = requests.get(rec, timeout=500)
```
dchiller, line 50: Same thing re timeout

musicbrainz/csv/get_genre.py
```diff
@@ -0,0 +1,64 @@
...
+
+    soup = BeautifulSoup(resp_wiki.text, "html.parser")
+    wikidata_row = soup.find("th", string="Wikidata:")
+    wikidata_value = ""
+    if wikidata_row:
+        wikidata_value = wikidata_row.find_next_sibling("td").find("a").text
+        relations_wiki.append("http://www.wikidata.org/entity/" + wikidata_value)
+    else:
+        relations_wiki.append("")
```
dchiller, line 59:
```suggestion
        relations_wiki.append(wikidata_value)
```

musicbrainz/csv/get_genre.py
Yueqiao12Zhang: > They want an informative User-Agent header so they can contact us if something is going wrong.
I still don't understand what is an User-Agent header.
> Read the information about rate limiting based on source IP. Have we constructed this to be within those limits?
They said the maximum rate is 50 request per second, I wait for 0.02s each for loop iteration. In case an HTTPError occurs, I wait for an longer time.

> Source IP address
> Unless you have agreed otherwise with MusicBrainz, the rule is as follows:
> 
> The rate at which your IP address is making requests is measured. If that rate is too high, all your requests will be declined (http 503) until the rate drops again. Currently that rate is (on average) 1 request per second.
> 
> For example: if your requests are coming in at 4 requests per second, we don't honour 25% of them and decline the other 75% - we decline 100% of them, until the rate drops to 1 per second or lower.

This is on the website. What does it mean to have agreed with MusicBrainz?

> If they reach the global limit, how does this script respond?
As long as they returns a HTTPError, the script drops the rate by waiting a longer time for each iteration.

dchiller: > I still don't understand what is an User-Agent header.

Happy to help! What do you know about it and what is confusing?

> They said the maximum rate is 50 request per second

This is incorrect. The part you quote does not say 50 requests per second. It says "(on average) 1 request per second". 

> What does it mean to have agreed with MusicBrainz?

It says "unless you have agreed with MusicBrainz". We don't have any agreement that I know of with MusicBrainz, so I think we are covered by their generic rules: "Rate is (on average) 1 request per second".

dchiller: > As long as they returns a HTTPError, the script drops the rate by waiting a longer time for each iteration.

Ok

musicbrainz/csv/get_genre.py
```diff
@@ -0,0 +1,64 @@
...
+        try:
+            resp = requests.get(url=URL, params=PARAMS, timeout=50)
+            resp.raise_for_status()
+            break
+
+        except HTTPError as exc:
+            code = exc.response.status_code
+
+            if code == 503:
+                if j == 2:
```
dchiller, line 30: If I set `MAX_REQUEST_RETRIES` to 2...what happens here?

musicbrainz/csv/get_genre.py
Yueqiao12Zhang: > I still don't understand what is an User-Agent header.

What is it's format and how do we pass it to the MusicBrainz API?

musicbrainz/csv/get_genre.py
Yueqiao12Zhang: Ok, this is a mystery number, I will fix it.

musicbrainz/csv/get_genre.py
```diff
@@ -0,0 +1,64 @@
...
+                    raise
+                time.sleep(j + 1)
+                continue
+
+            raise
```
dchiller, line 35: What's this doing here? I think you could get rid of `raise` here and `break` on line 24.

musicbrainz/csv/get_genre.py
```diff
@@ -0,0 +1,64 @@
...
+
+    temp = resp.json()
```
dchiller, line 37: This could use a better variable name...what does `temp` mean?

musicbrainz/csv/get_genre.py
```diff
@@ -0,0 +1,64 @@
...
+    data += temp["genres"]
+    time.sleep(0.02)
```
dchiller, line 39: Magic number...needs explanation.

musicbrainz/csv/get_genre.py
dchiller: The API documentation page gives a nice suggested format for the User-Agent header:

image

See what you can find in the requests package documentation about how to pass a request header as part of your request.

musicbrainz/csv/get_genre.py
Yueqiao12Zhang: This is because HTTPError is not only when the server returns code 503, when it is unavailable. If other codes are returned and it is a HTTPError, then we do not handle it and raise that exception directly.

musicbrainz/csv/get_genre.py
```diff
@@ -0,0 +1,64 @@
...
+"""
+
+import time
+import requests
+from requests.exceptions import HTTPError
+import pandas as pd
+from bs4 import BeautifulSoup
+
+URL = "https://musicbrainz.org/ws/2/genre/all"
+PARAMS = dict(fmt="json", limit="50")
```
dchiller, line 12:
```suggestion
PARAMS = {"fmt"="json", "limit"="50"}
```

musicbrainz/csv/get_genre.py
```diff
@@ -0,0 +1,64 @@
...
+MAX_REQUEST_RETRIES = 3
+
+data = []
+max_records = requests.get(url=URL, params=PARAMS, timeout=50).json()["genre-count"]
+
+for i in range(0, max_records, 50):
+    PARAMS["offset"] = i
```
dchiller, line 19:
```suggestion
    PARAMS["offset"] = str(i)
```

Yueqiao12Zhang: Why is it a string? Should it be an Int?

ahankinson: URL parameters are always strings.

musicbrainz/csv/get_genre.py
dchiller: Headers look great!

dchiller (approved):

PR #115, opened 2024-07-29
Pyproject location change

Yueqiao12Zhang:
The pyproject.toml was only in the csv2rdf folder. It should be global to all folders in the linkedmusic-datalake folder.


dchiller:
This would be a good place for a more specific commit messages!

- "change location" (of what? to where?)
- "upgrade modules and packages" (what modules? what packages? where?)
- "update according to warning" (what are you updating? according to what warning?)

Yueqiao12Zhang:
Change location: of pyproject.toml and poetry.lock from csv2rdf to the main directory
upgrade modules: poetry and validators
warning: I use poetry for dependency management only, so I disable the package mode.

dchiller:
> I've learned that commit messages should be as short as possible, so that's why I simplify them.

Not really...

They should succinctly describe what happened in the commit. In short, you want to be able to figure out roughly what the commit changed and why in the message.

dchiller:
You can git rebase here with the "reword" keyword to fix these.

`git rebase -i main`

Then, replace `pick` with `reword` or `r` in front of each commit. This will let you change the message.

dchiller:
I think there's now an errant csv file in this PR...

dchiller:
The commit history here is still weird...there are a bunch of duplicated commits... Did you rebase?

Yueqiao12Zhang:
> The commit history here is still weird...there are a bunch of duplicated commits... Did you rebase?

I renamed the commits, and it was like this.

dchiller:
I just rebased this branch off of `main` and achieved the following commit history:

image

So I'm not sure what you did, but perhaps try again?

dchiller (approved):

PR #117, opened 2024-08-01
feat: reads the old json if it exists

Yueqiao12Zhang:
If the JSON file already exists and is filled with mapping values, then the script reads it and append additional predicates to the file.

### DIFF
--- a/csv2rdf/get_relations.py
+++ b/csv2rdf/get_relations.py
@@ -5,9 +5,12 @@ It reads a mapping.json file and all .csv files in that folder.
 For each CSV file provided as input, the script performs the following steps:
 
 1.  Read folder: If a mapping file already exists, 
-then the script reads that file and append any additional predicates.
+then the script reads that file and saves any predicate mappings that exist in that file.
 2.  Extract Headers: The script extracts the headers from each CSV file.
-3.  Write to JSON: The script writes these headers to a JSON file,
+3.  Write to JSON: If a column header extracted from a CSV file was mapped
+in the existing mapping file, keep the existing mapping. If we find a header that
+did not exist in the existing mapping file, create a key for that header with a blank
+predicate value.  The script writes the resulting mapping to a JSON file,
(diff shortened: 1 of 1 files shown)


csv2rdf/get_relations.py
dchiller, line 1: Check the module docstring...is it up-to-date with the changes you made in this commit?

dchiller (changes requested):

csv2rdf/get_relations.py
```diff
@@ -1,10 +1,14 @@
...
 
 For each CSV file provided as input, the script performs the following steps:
 
-1. Extract Headers: The script extracts the headers from each CSV file.
-2. Write to JSON: The script writes these headers to a JSON file, which will serve as the mapping file for the csv2rdf_single_subject.py script.
+1.  Read folder: If a mapping file already exists, 
+then the script reads that file and append any additional predicates.
+2.  Extract Headers: The script extracts the headers from each CSV file.
+3.  Write to JSON: The script writes these headers to a JSON file,
+which will serve as the mapping file for the csv2rdf_single_subject.py script.
```
dchiller, line 11:
```suggestion
1.  Read folder: If a mapping file already exists, 
then the script reads that file and saves any predicate mappings that exist in that file.
2.  Extract Headers: The script extracts the headers from each CSV file.
3.  Write to JSON: If a column header extracted from a CSV file was mapped
in the existing mapping file, keep the existing mapping. If we find a header that
did not exist in the existing mapping file, create a key for that header with a blank
predicate value.  The script writes the resulting mapping to a JSON file,
which will serve as the mapping file for the csv2rdf_single_subject.py script.
```

dchiller (approved):
I made a suggestion for part of the docstring...do what you will with that and then this looks merge-ready.

Yueqiao12Zhang:
> I made a suggestion for part of the docstring...do what you will with that and then this looks merge-ready.

Thank you very much!

PR #119, opened 2024-08-07
The session correction

Yueqiao12Zhang:
Make changes according to suggested changes.


csv2rdf/thesession/mapping.json
```diff
@@ -37,13 +36,13 @@
...
     "setting_id": "https://thesession.org/setting_id",
-    "type": "http://www.wikidata.org/prop/direct/P2308",
+    "type": "http://www.wikidata.org/prop/direct/P136",
     "type_wiki": "http://www.wikidata.org/prop/direct/P2308",
     "meter": "http://www.wikidata.org/prop/direct/P3440",
     "meter_wiki": "http://www.wikidata.org/prop/direct/P3440",
     "mode": "http://www.wikidata.org/prop/direct/P826",
     "mode_wiki": "http://www.wikidata.org/prop/direct/P826",
-    "tunebooks": "http://www.wikidata.org/entity/Q66382988",
+    "tunebooks": "https://thesession.org/tunebooks",
```
dchiller, line 45: I don't think this is correct. This should be a predicate, right?

csv2rdf/thesession/mapping.json
dchiller, line 1: I'm actually confused by all the mappings that have "_id"?

csv2rdf/thesession/mapping.json
Yueqiao12Zhang: It is a predicate, that's why I changed the wiki Q value.

csv2rdf/thesession/mapping.json
Yueqiao12Zhang: What is the confusion?

dchiller: Take, for example, `settings_id`. You've given it this predicate: https://thesession.org/setting_id.

Can you say in words what that predicate means?

Yueqiao12Zhang: They are indeed confusing, since they are entities instead of properties. I change them in #125.

dchiller: I don't think this is affected by #125, because, for example, `settings_id` is not the first column of a dump, right? It's like the 4th or 5th column in `sets`

dchiller: If you aren't worried about the reconciliation now, that's fine. We can merge this PR and address later.

fujinaga (commented):
Can you comment what you are doing with this code and why.

Yueqiao12Zhang:
> Can you comment what you are doing with this code and why.

#134 #135

PR #125, opened 2024-08-12
Get relations.py remove first column property

Yueqiao12Zhang:
Remove the first column as a property during the generation of mapping.json, since the first column is not a property. 
Also clean up the conversion process by removing the unused objects in the old mapping.json.

### DIFF
--- a/csv2rdf/get_relations.py
+++ b/csv2rdf/get_relations.py
@@ -28,11 +28,9 @@ try:
     with open(OUTPUT_NAME, "r", encoding="utf-8") as mapping:
         dt = json.load(mapping)
 except FileNotFoundError:
-    dt = {}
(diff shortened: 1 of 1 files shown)


csv2rdf/get_relations.py
```diff
@@ -30,15 +30,20 @@
 except FileNotFoundError:
     dt = {}
     dt["entity_type"] = []
+    
+dt_out = {}
+dt_out["entity_type"] = dt["entity_type"]
```
dchiller, line 35:
```suggestion
    dt = {"entity_type" : []}
    
dt_out = {"entity_type" : dt["entity_type"]}
```

dchiller (approved):
Made one suggestion if you want...

Looks good!

PR #126, opened 2024-08-12
Csv2rdf entity_type order-independent

Yueqiao12Zhang:
Now the entity_type object is no longer a list. It's changed to a dict that maps the entity type from Wikidata to the filename.


dchiller (approved):
Nice... link the issue and I'm good.

PR #129, opened 2024-08-13
Musicbrainz fetch files script

Yueqiao12Zhang:
Create fetch.py and untar.py to download and unzip the real data sets from MusicBrainz.
Also modify the documents accordingly.


musicbrainz/csv/untar.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def extract_file(folderpath, dest_folder):
+    """
+    untar the downloaded files
+    """
+    for filepath in glob.glob(f"{folderpath}/*.tar.xz", recursive=False):
+        with tarfile.open(f"{filepath}", "r:xz") as tar:
+            for member in tar.getmembers():
+                if member.name.startswith("mbdump"):
```
dchiller, line 16: Could you comment something about why `mbdump`? Are there other members? Why aren't they wanted? Just a line or two about that either here or in the module docstring.

musicbrainz/csv/fetch.py
```diff
@@ -0,0 +1,59 @@
...
+    response = requests.get(url, timeout=50)
+    soup = BeautifulSoup(response.text, "html.parser")
+
+    # Extract all directories
+    dirs = [a.text for a in soup.find_all("a") if a.text.startswith("latest-is-")]
+
+    latest_dir = dirs[-1]
+    latest_dir = latest_dir[10:]
+    resp_url = url + latest_dir + "/"
+    return resp_url
```
dchiller, line 25: You could get this more simply and without all of the magic number string manipulation with:

`requests.get("https://data.metabrainz.org/pub/musicbrainz/data/json-dumps/LATEST").text.strip()`

musicbrainz/csv/fetch.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+    API calls to fetch all .tar.xz files
+    """
+    response = requests.get(url, timeout=50)
+    soup = BeautifulSoup(response.content, "html.parser")
+
+    # Find all links ending with .tar.xz
+    tar_xz_files = [
+        url + a["href"] for a in soup.find_all("a") if a["href"].endswith(".tar.xz")
+    ]
```
dchiller, line 38: I think you should probably get these files by name, rather than grab all of them.

Yueqiao12Zhang: I don't understand what you mean by getting them by name. Are you saying getting them by names like "recording.tar.xz"?

dchiller: Yes. Like, have a list of names of the files you are trying to download and run through that list.

musicbrainz/csv/untar.py
```diff
@@ -0,0 +1,25 @@
...
+
+def extract_file(folderpath, dest_folder):
+    """
+    untar the downloaded files
+    """
+    for filepath in glob.glob(f"{folderpath}/*.tar.xz", recursive=False):
+        with tarfile.open(f"{filepath}", "r:xz") as tar:
+            for member in tar.getmembers():
+                # in the tar file, we only need the dumps in this folder.
+                # Other files contains unecessary info.
```
dchiller, line 17:
```suggestion
                # Other files contains unnecessary info.
```

dchiller:
Did you mean to delete the files in `data/raw`?

musicbrainz/csv/fetch.py
Yueqiao12Zhang: Ok. I will change this.

Yueqiao12Zhang:
> Did you mean to delete the files in `data/raw`?

Yes, since we can fetch the original files now, we don't need the shortened files now.

dchiller (approved):

PR #141, opened 2024-08-14
relations.py output mapping.json sort key feature

Yueqiao12Zhang:
Add the sort key output feature to get_relations.py, plus apply this to all the output test mappings.json file associated with this change.


dchiller (approved):
Thanks, nice!

PR #166, opened 2024-08-23
Namespace binding

Yueqiao12Zhang:
Binds the WDT and WD namespaces to the generation of RDF graph.


dchiller (approved):
Ok! 

But no more commit messages that are "update according to changes" :wink:

Issue #67, opened 2024-06-13
Reconciling the raw SQL from Cantus to CSV


Yueqiao12Zhang:
I have an SQL data dump from CantusDB. From line 2907, the TSV data of chants starts. I copied 7000 records from that line and started a reconciliation on it.

Yueqiao12Zhang:
In the SQL data dumps, almost all links are represented in a Cantus DB ID, e.g. feast_id, genre_id, office_id, etc. It can be appended to a URL such as "https://cantusdatabase.org/feast/2219". Can I get the name/title of these links directly in OpenRefine?

fujinaga:
I don’t understand your question. Can you elaborate?

Yueqiao12Zhang:
For example, if we have a feast for an entity, the only available information in the CSV is 2219. If I want to reconcile it, I need to get the website's title "https://cantusdatabase.org/feast/2219", which is Gregorii. So I'm guessing that I might need to request the title of the https.

Yueqiao12Zhang:
What is an office of a chant?
https://cantusdatabase.org/office/4151

Yueqiao12Zhang:
For a feast, is Dom. in Palmis its name or is it Palm Sunday?

dchiller:
> What is an office of a chant?

The office of a chant is what service during the day a chant was sung at (see https://cantusdatabase.org/offices/) (I think this would be the Wikidata entry: https://www.wikidata.org/wiki/Q946731). 

> For a feast, is Dom. in Palmis its name or is it Palm Sunday?

Both! They refer to the same day but in different languages (Latin v English)

dchiller:
> For example, if we have a feast for an entity, the only available information in the CSV is 2219. If I want to reconcile it, I need to get the website's title "https://cantusdatabase.org/feast/2219", which is Gregorii. So I'm guessing that I might need to request the title of the https.

If we want to reconcile feasts in CantusDB with Wikidata, you are going to need something besides the CantusDB ID for that feast, so I'm not sure what the "title of the https" is, but you certainly are going to want "Gregorii" as opposed to "2219".

Yueqiao12Zhang:
> > For example, if we have a feast for an entity, the only available information in the CSV is 2219. If I want to reconcile it, I need to get the website's title "https://cantusdatabase.org/feast/2219", which is Gregorii. So I'm guessing that I might need to request the title of the https.
> 
[...]

PR #265, opened 2025-05-13
Reconcile CantusDB

### LINKED ISSUE #248: Reconcile CantusDB with Wikidata (following Yueqiao's existing documentation)

TTG3333:
The reconciliation process uses the full sources downloaded from the database website, instead of a database dump. This makes it easier to perform updates to the retrieved data.
- Clarified the existence of a test source file to not download the entire database when testing
- Re-ran the reconciliation process on the whole database and updated the history and exports files to make future reconciliation efforts easier
- Updated the documentation to make it clearer and to reflect changes
- Removed large data files from the repository and added gitignores to prevent them from being added again

Resolves #248 and #67


liampond (approved):
Great! Makes sense

PR #271, opened 2025-05-15
Small fix to The Session documentation

linneakirby:
Edit one line of The Session reconcile procedures file and update the corresponding history JSON file.


liampond (approved):
Looks good!

PR #280, opened 2025-05-20
DIAMM Reconciliation process

### LINKED ISSUE #258: Reconcile DIAMM Database
- Figure out a way to export data or to get a database dump
- Convert the data to CSV to reconcile it
- Re-export as json-ld

TTG3333:
Entire reconciliation process for DIAMM, from fetching the data all the ay to converting it to RDF (turtle)
- Both synchronous and asynchronous crawler to fetch the database
- Script and documentation to convert the downloaded JSON to CSVs for reconciliation
- Reconciliation instructions and history and export files
- Script to convert the reconciled data into Turtle format
- Additions to the gitignores to remove large files from the repository and to prevent data files from getting uploaded

Resolves #258


pyproject.toml
```diff
@@ -15,8 +15,14 @@ urllib3 = "^2.2.2"
 beautifulsoup4 = "^4.12.3"
 requests = "^2.32.3"
 tqdm = "^4.67.1"
+psycopg2-binary = "^2.9.10"
 
 
+[tool.poetry.group.dev.dependencies]
+aiohttp = "^3.11.18"
```
dchiller, line 22: Are these really dev dependencies?

code/diamm/async_crawler.py
```diff
@@ -0,0 +1,190 @@
...
+    "regions",
+]
+
+visited = set() # tuple (str, str) representing (type, id), example is ("compositions", "117")
+
+to_visit = [("sources", "117")]  # Starting point
+
+def print_stdout_stderr(message):
+    """
+    Prints a message to both stdout and stderr.
```
dchiller, line 51: Can you help me with an example here? 

In general, I think using the `logging` module is best for logging, in part because it built-in can do this kind of "direct some level of messages here and others there" kind of thing.

pyproject.toml
TTG3333: No they aren't, I moved aiohttp and aiofiles to be in the main dependancy group

code/diamm/async_crawler.py
TTG3333: The main usecase is because the script is written with the assumption that the stdout will be sent to a file (ie running `python ./async_crawler.py > log.txt`), but you still might want to see certain messages (like crawlers crashing or finishing), so they need to get sent to both stdout and stderr to show up in the log file and the terminal.

I'll take a look at the `logging` module to see if I can implement it with that instead.

ahankinson:
How many requests per second does this script fire? What are you doing to prevent overloading the DIAMM website?

TTG3333:
> How many requests per second does this script fire? What are you doing to prevent overloading the DIAMM website?

It fires up to 40 per second, I added a rate limiter to it yesterday.

liampond (approved):

code/diamm/async_crawler.py
dchiller, line 1: Was this file formatted?

code/diamm/async_crawler.py
```diff
@@ -0,0 +1,209 @@
...
+"""
+
+import os
+import asyncio
+import aiohttp
+import aiofiles
+import json
+import re
+import logging
+from aiolimiter import AsyncLimiter
```
dchiller, line 19: Import order in this and other py files don't currently conform to PEP 8 standard. https://peps.python.org/pep-0008/#imports

A linter will catch this...

code/diamm/async_crawler.py
```diff
@@ -0,0 +1,209 @@
...
+# Console handler for logging to stderr
+console_handler = logging.StreamHandler()
+console_handler.setLevel(logging.ERROR)
+console_format = logging.Formatter("%(levelname)s - %(message)s")
+console_handler.setFormatter(console_format)
+logger.addHandler(console_handler)
+
+visited = set() # tuple (str, str) representing (type, id), example is ("compositions", "117")
+
+to_visit = [("sources", "117")]  # Starting point
```
dchiller, line 71: Can you document why this was chosen?

dchiller (commented):
A general question I have is about the decision to use a scraper? If I look at the issue, we potentially had access to the contents of the database straight-away? 

Since DIAMM is a partner in this project, my assumption would be coordinating with them in some sort of export process is more efficient. But there are maybe reasons against this in this case? Could those be documented (perhaps in the issue, since the last mention there is of a SQL dump)?

code/diamm/async_crawler.py
TTG3333: I did, but I made changes since and forgot to reformat, I just fixed that and the PEP 8 imports

code/diamm/async_crawler.py
TTG3333: Documentation added to the code, essentially it was the example given on DIAMM's website, and when I tried it out, it looked like it allowed m to reach pretty much all the database, so I stuck with it

dchiller: Gotcha... thanks.

In that case, we need to open an issue (out of scope for this PR) to check this for completeness.

TTG3333: Issue created #285

doc/diamm/README.md
```diff
@@ -0,0 +1,33 @@
+# DIAMM Database Reconciliation Documentation
+
+## 1. Fetching Data
+
+The website supports content negotiation, either by adding `?format=json` to the end of the URL or by sending the HTTP header `Accept: application/json`. To retrieve the data, we use a web crawler, either `code/diamm/crawler.py` or `code/diamm/async_crawler.py` for a much faster one. The data is sent to `data/diamm/raw/<type>/<pk>.json`
+
+I chose to not download the pages for cities, countries and regions because we can easily reconcile against Wikidata for that. I also chose to not download the pages on authors because they do not contain anything of use at all.
```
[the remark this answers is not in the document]
TTG3333: I adde the documentation to the readme

TTG3333:
For rate limiting, I reduced it to 2 concurrent visits and 10 requests per second

TTG3333:
> A general question I have is about the decision to use a scraper? If I look at the issue, we potentially had access to the contents of the database straight-away?
> 
> Since DIAMM is a partner in this project, my assumption would be coordinating with them in some sort of export process is more efficient. But there are maybe reasons against this in this case? Could those be documented (perhaps in the issue, since the last mention there is of a SQL dump)?

My understanding of #258 is that getting the SQL dump would be harder than using a scraper, thus making it harder to keep our copy of the database up-to-date.

dchiller:
> My understanding of #258 is that getting the SQL dump would be harder than using a scraper, thus making it harder to keep our copy of the database up-to-date.

To me, a regular SQL export of database contents is a much easier way to keep us up to date than running a scraper. Many fewer points of potential failure. 

Happy to make this another issue if that makes sense. I'm gonna go ask this in #258 though just so we can get that thought process written down.

doc/diamm/README.md
```diff
@@ -0,0 +1,35 @@
+# DIAMM Database Reconciliation Documentation
+
+## 1. Fetching Data
+
+The website supports content negotiation, either by adding `?format=json` to the end of the URL or by sending the HTTP header `Accept: application/json`. To retrieve the data, we use a web crawler, either `code/diamm/crawler.py` or `code/diamm/async_crawler.py` for a much faster one. The data is sent to `data/diamm/raw/<type>/<pk>.json`
+
+I chose to not download the pages for cities, countries and regions because we can easily reconcile against Wikidata for that. The only information that they contain is the list of archives, sources, and organizations in that city/country/region, which we already have because archives, sources, and organizations also have a field indicating which city/country they're in.
+
+I also chose to not download the pages on authors because they do not contain anything of use at all in terms of classifying or adding data to the other data types in the database. The "author" data type seems to represent (modern) people who wrote papers that reference or mention various institutions or sources.
```
dchiller, line 9: Thanks for this explanation. The bibliographic information on those pages seems quite relevant/interesting/useful!

Fine if we want to bracket this for now and make an issue for it -- I'm wondering how in the current method we capture bibliographic relations between say, a source and a paper and/or author.

TTG3333: I believe that the current process skips authors and bibliographic relations, but that is something that can (relatively) easily be added

doc/diamm/README.md
```diff
@@ -0,0 +1,35 @@
+# DIAMM Database Reconciliation Documentation
+
+## 1. Fetching Data
+
+The website supports content negotiation, either by adding `?format=json` to the end of the URL or by sending the HTTP header `Accept: application/json`. To retrieve the data, we use a web crawler, either `code/diamm/crawler.py` or `code/diamm/async_crawler.py` for a much faster one. The data is sent to `data/diamm/raw/<type>/<pk>.json`
+
+I chose to not download the pages for cities, countries and regions because we can easily reconcile against Wikidata for that. The only information that they contain is the list of archives, sources, and organizations in that city/country/region, which we already have because archives, sources, and organizations also have a field indicating which city/country they're in.
```
dchiller, line 7: Again, thanks for adding this. 

My question is: when I'm reconciling DIAMM with Wikidata, aren't I interested in the fact that https://www.diamm.ac.uk/cities/694/ and https://www.wikidata.org/wiki/Q42462 are the same?

Maybe one need not scrape the cities page, but if I look at how you are processing the archives entities, you are taking the city name, and missing out on the DIAMM URI, right?

doc/diamm/README.md
dchiller: Gotcha, ok -- don't want to initiate scope creep. #287 created.

doc/diamm/README.md
TTG3333: Yes, I take the city/country name and reconcile that, but I don't at any point store or use the DIAMM URI for the city/country

dchiller: I realize you are just trying to re-implement an existing process, so if this isn't in the original, I'll make a new issue.

TTG3333: There was no existing process for DIAMM, it wasn't one that was worked on before.

dchiller: I see...ok. In that case, I leave it up to you if you want to implement that here or make an issue.

TTG3333:
@dchiller Am I good to merge this PR?

dchiller (approved):

PR #282, opened 2025-05-21
Update the jsonld context for SIMSSADB

### LINKED ISSUE #278: Update context.jsonld for SIMSSADB
When the jsonld process was modified for SIMSSADB, the context was never modified to reflect the modified fields.

TTG3333:
- Mapped the fields that were previously added to wikidata properties in the context.jsonld file
- Also changed the link to the context to link directly to the DDMAL repository, and changed the file path to be accurate

Resolves #278


jsonld_approach/simssadb/jsonld/context.jsonld
```diff
@@ -4,9 +4,10 @@
         "mw": "https://db.simssa.ca/musicalworks/",
         "simssadb_file": "https://db.simssa.ca/files/",
         "musical_work_variant_titles": "wdt:P1476",
-        "contributor": "wdt:P767",
-        "contributor_role": "wdt:P3831",
-        "contributor_name":"wdt:P2561",
+        "composer": "wdt:P86",
```
dchiller, line 7: nice correction

jsonld_approach/simssadb/jsonld/context.jsonld
```diff
@@ -4,9 +4,10 @@
...
+        "composer_name":"wdt:P2561",
```
dchiller: what's the subject of the triple that has this property?

TTG3333: The subject is the musical work for composer, and the subject for composer_name is the composer, which is why it uses 2561 (name)

liampond (approved):

dchiller (approved):

Issue #407, opened 2025-07-02
Add an RDF:type triple to each entity


TTG3333:
Currently, when we want to perform a query like "select all archives in DIAMM that fulfill x condition", the only way for us to separate archives from other entities in DIAMM is to perform string filtering on the URI of entities to ensure that they begin with `https://www.diamm.ac.uk/archives/`.

Following a discussion during today's meeting, we decided to add a triple with the `RDF:type` predicate pointing to a URI on the linkedmusic.ca website that would point to a URI for each entity type, like `https://linkedmusic.ca/graphs/diamm/archive`, following and building on the URI structure established in #373

TTG3333:
I will use singular as grammatically it makes more sense

TTG3333:
Fair enough, although if I approach it from a prefix/SPARQL query point of view, currently we have the `diamm:` prefix, which expands to `https://linkedmusic.ca/graphs/diamm/`, and as such we can do `diamm:Archive` to get the archive entity type.

We could always make a second prefix (`diamm2:`?) that points to `https://linkedmusic.ca/diamm/` but then we'd have 2 prefixes for the same database.

What about having the entity reside at `https://linkedmusic.ca/diamm/Archive` like you suggested but also changing the graph to either be `https://linkedmusic.ca/diamm/` or `https://linkedmusic.ca/diamm/graph` (I personally think that the first is better) so that we can use the same prefix for both?

PR #420, opened 2025-07-09
Add an rdf type triple to MusicBrainz and DIAMM

TTG3333:
This PR adds an RDF:type triple to give an entity type for each entity in DIAMM And MusicBrainz, following discussion in #407 . It doesn't close it as the other databases handled by the general RDF conversion script will be done in another PR by @SCN-MNG


liampond (approved):
Looks good!

PR #438, opened 2025-08-12
Merge global jukebox rdf conversion branch

linneakirby:
Adds toml file for converting the global jukebox to rdf


liampond (approved):
Looks good

